PR #10, opened 2019-09-18
add an admin folder with a roadmap

### LINKED ISSUE #9: Add a plan
I think we need a document to describe, for each chapter
* their learning objectives
* their prerequisites
* what other sections they are prerequisites of

This would be mostly for contributors to look at, so they can more easily see where new content could fit in.

quantumjim:
closes #9


quantumjim:
Probably we should do the this by adding metadata to sections and building the dependency graph programmatically. But this'll do for now.

Issue #4, opened 2019-09-11
Grover's algorithm


ozlemsalehi:
In the Grover's algorithm, the initial state is named both s and \psi_t. I think it is a bit confusing. Also, the unitary operator in the picture is named U_w whereas it is named U_f in the text. Somewhere below you define \psi_0 and then define \psi_t as the state reached after t steps but you have already used \psi_t as the name for the initial state.

Geometrically this corresponds to a reflection of the state  |ψt⟩  about  −|w⟩.
Isn't it a reflection about |s'> ? 

The case of more than one solutions can be explained in more detail.

Your choice of oracle is misleading since it is very common to think that you should know in advance which elements are marked to apply the oracle. Instead, it would be better to use phase kickback for the oracle part.

aasfaw:
Thanks @ozlemsalehi. Would you like to open a pull request with these changes?

frankharkins:
Fixed in PR#178

PR #21, opened 2019-09-30
Grover's Algorithm, inconsistency in notation of the reflection operator

MadhavJivrajani:
In the images, reflection operator is denoted by U_w but in the text it is denoted by U_f


aasfaw:
Partially fixes #4

Issue #19, opened 2019-09-23
Explore migrating to Jupyter Book


delapuente:
See if it requires a Jupyter hub or not, or when.

delapuente:
See my [`jupyter-book` branch](https://github.com/delapuente/qiskit-textbook/tree/jupyter-book) for advances on this.

delapuente:
See https://delapuente.github.io/qiskit-textbook for a live version.

Not merging yet, since we need to wait for #20 and ensure Travis fails when some notebook does not run.

PR #37, opened 2019-10-10
Navigation edit

starktech23:
Adding a very early stage QoL feature to navigate between the chapter pages without having to go back to the index page each time similar to a textbook UI. Template followed similar to index page. Only previous page, next page, and index page links on each chapter.

### DIFF
--- a/ch-algorithms/bernstein-vazirani.ipynb
+++ b/ch-algorithms/bernstein-vazirani.ipynb
@@ -446,6 +446,20 @@
     "import qiskit\n",
     "qiskit.__qiskit_version__"
    ]
+  },
+  {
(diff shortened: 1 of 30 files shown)


starktech23:
Future navigation UI may include moving to any chapter page and not just adjacent ones with something like a sidebar. Not sure how to implement that with Jupyter Notebooks yet.

aasfaw:
Hi @starktech23 -- this is great. We are working on integrating the textbook into Jupyter book. See issue #19 for a sneak peek :) I think you'll be very happy.

starktech23:
@aasfaw very happy indeed! 19 looks great, thanks for the reference :D

PR #40, opened 2019-10-10
Added two LaTex figures, fixed couple typos

stanescuUW:
Hopefully you find this helpful.


stanescuUW:
I can do the conversion but the image becomes quite low quality and blurry; not fit for a textbook IMO. Will do the name changes and insert figures in the couple other places, but wait for your input on the conversion. We really won't gain too much in size but lose a lot in quality.

aasfaw:
Thanks for clarifying @stanescuUW. That sounds good -- let's do pdfs for now.

stanescuUW:
OK now. I'm new with GitHub, so I'm not sure I did everything but I did push the changes to my branch. I understand Git takes it from here. Let me know if anything else, thx!

frankharkins:
Hi @stanescuUW, thanks very much for your contribution! The pdf files are not rendering for me when I build, are you able to export them to SVG instead? This should maintain quality while keeping the file size small. 

Also I think you are accidentally pushing some files not meant for the textbook, for example: `firstNotebook-checkpoint.ipynb`. You can see this in the "Files Changed" tab at the top of this page.

stanescuUW:
@frankharkins Hi Frank, you are talking to a total GitHub newbie. I really do not know yet how to process your change request. While it is very surprising that you can't process the pdf files, they should be easy to convert to svg and used as such. Also, I do have the figures, but I may have deleted the branch for this pull request, so I have no way to go back and process it. But what matters is for these figures to find their way in the book if you think they're useful - not for my pull request to be merged. Also, can't you guys on that end remove the extraneous files and just pull the necessary ones?

frankharkins:
@stanescuUW that's okay, I'll try and sort it out on Monday. For future reference though, you should be able to click "compare" before you submit a request and you can see which files you are sending. You can then remove any you don't want, commit again then do the pull request. Have a nice weekend!

PR #48, opened 2019-10-20
Fix in equation of Matrix Exponential

nicosquare:
When replacing the calculated values in the decomposition of the Involutory Matrix, you forgot to remove the replaced element in the sin part of the equation, it is just necessary to remove the factor from the equation before equalizing to sin and cos.


frankharkins:
Hi, thanks for your contribution! This issue was resolved in PR#58.

PR #49, opened 2019-10-22
Fix incorrect explanation of Deutsch-Jozsa algorithm.

dlyongemallo:
The qiskit code will also need to be fixed to match the description, which I haven't done.

### DIFF
--- a/ch-algorithms/deutsch-josza.ipynb
+++ b/ch-algorithms/deutsch-josza.ipynb
@@ -63,7 +63,7 @@
     "We are given a hidden Boolean function $f$, which takes as as input a string of bits, and returns either $0$ or $1$, that is \n",
     "<center>$f(\\{x_0,x_1,x_2,...\\}) \\rightarrow 0 \\textrm{ or } 1 \\textrm{ , where } x_n \\textrm{ is } 0 \\textrm{ or } 1$.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


attp:
Hi @dlyongemallo ! Thanks for going through this chapter of the textbook in such detail.

I've incorporated some of your comments in my pull request #51 . It was unclear when the text was referring to input bitstring combinations, rather than input bitstring and output bit combinations. I hope it makes more sense now.

I've left the example with a=3 (11 in binary), as it makes the maths and circuit simpler, with only 3 qubits required; rather than an a=6 (110 in binary) as you suggest, since that'll mean 4 qubits.

dlyongemallo:
You misunderstanding what I'm referring to with 0110: the example _already_ corresponds to the bitstring 0110. For an $n$-qubit Deutsch-Jozsa problem, the balanced functions are represented by bitstrings which are $2^n$ bits long. For example, for $n = 2$, you have to specify the 4 values of $f(00)$, $f(01)$, $f(10)$, and $f(11)$.

The explanation in the current state doesn't match the example. It _happens by coincidence_ that taking the bitwise product with $11$ produces the outcome $f(00)f(01)f(10)f(11) = 0110$, but you will not be able to represent all balanced functions with a bitwise product in general. This becomes clear for $n > 2$, since there are $\binom {2^{n}} {2^{n-1}}$ balanced functions but only $2^{n}$ bitstrings.

Just consider $n = 3$. There are $\binom {8} {4} = 70$ balanced functions, so how can $2^3 = 8$ bitstrings represent all of them? For $n = 4$, there are $\binom {16} {8} = 12870$ balanced functions, and only $2^4 = 16$ bitstrings. And so on.

Issue #54, opened 2019-10-23
Issues with Deutsch-Josza


quantumjim:
Some issues have been raised with the section on the Deutsch-Jozsa algorithm. See PR #49 and [these details](https://twitter.com/dlyongemallo/status/1186762479512248322) on Twitter.

attp:
Also PR #52

frankharkins:
Fixed in PR#177

PR #55, opened 2019-10-23
Fix Deutsch-Josza section

attp:
Corrected text to access concerns summarised in Issue #54 .

### DIFF
--- a/ch-algorithms/deutsch-josza.ipynb
+++ b/ch-algorithms/deutsch-josza.ipynb
@@ -165,7 +165,7 @@
     "\n",
     "For a balanced function, it is more complicated:\n",
     "\n",
-    "$\\qquad$ There are $2^{n}-1$ different hidden bitstring configurations for an $n$-bit balanced function oracle. These can be defined by one of the bitstrings from $1$ to $2^n-1$ inclusive. Given a particular hidden bitstring, $a$, the oracle is the bitwise product of $x$ and $a$, which is implemented as a multi-qubit f-controlled-NOT gate with the second register, as per Reference [2]. "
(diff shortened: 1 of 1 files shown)


attp:
@aasfaw, @quantumjim, @dlyongemallo , please review and comment. am trying just to change the text and not the code, if possible.

dlyongemallo:
In my opinion, the code needs to be changed. I can see that making a function to generate a random balanced function for n > 3 may introduce complexity to the code that takes away from the focus on the Deutsch-Jozsa algorithm itself. To keep things simple, you can make it clear that the example only works for n = 2. (This is how it's done in Jack Hidary's recently released textbook, for example.) And then even in that case, you need to modify the code to generate the other half of the balanced functions.

|oracle|f(00)|f(01)|f(10)|f(11)|
|-----|-----|-----|-----|-----|
|c=0|0|0|0|0|
|b=1|0|0|1|1|
|b=2|0|1|0|1|
|b=3|0|1|1|0|
|c=1|1|1|1|1|
|???|1|1|0|0|
|???|1|0|1|0|
|???|1|0|0|1|

Similar to how the circuit for half of the constant functions applies x(qr[n]) (line 297), the circuits for half of the balanced functions (the block after line 298) need to apply x(qr[n]). These correspond to the missing lines (marked "???") in the above table. The strings whose circuits have x(qr[n]) correspond to the Boolean complements of strings without it. This is why "0000" and "1111" are distinguished by the presence of this gate. As another example, the circuit for "1100" is the same as the circuit for "0011" (b=1) but with the x(qr[n]) gate. So either you should take out the qr[n] gate and say that you only care about functions for which f(00) = 0, or you need to put it in for the balanced functions. But it's weird to have both (all) constant functions but only half of the balanced functions.

I think the explanation as it is written, involving bitwise product with a fixed hidden bitstring, is misleading. The reason I came upon this notebook was because I was trying to explain the Deutsch-Jozsa algorithm to someone, and I think pedagogically, it needs to be emphasised that you're talking the bitwise product of x with f(x) which can in general depend on x. There are O(2^{2^{n}}) balanced functions you can choose from, but by restricting it to fixed hidden bitstrings, you're only picking out of O(2^n). If this simplifies things for an example, it's fine, but it needs to be made clear that it's a simplified version and not the full Deutsch-Jozsa problem. My concern is that people who have never seen the algorithm before will learn the wrong thing about it based on the simplification.

PR #59, opened 2019-10-28
Update calibrating qubits openpulse

sudhakarsq:
There were several issues with this file:
1. Usage of `pulse.DeviceSpecification.create_from(backend)` which is now deprecated. Replaced with `pulse.PulseChannelSpec.from_backend(backend)`.
2. Consequently, `backend_devicespec.q` instruction doesn't work. Replaced with `backend_devicespec.qubits`.
3. Also, `backend_devicespec.mem` calls don't work either. Replaced with `backend_devicespec.memoryslots`.
4. `backend.retrieve_job` calls with hardcoded jobIDs. Replaced with `job.job_id()` as an argument.
5. `job_monitor(job, monitor_async='True')` doesn't work since the second argument is not supported. Omitted the second argument since the next instruction anyways waits for the jobID from the last `backend.run` instruction.


sudhakarsq:
@aasfaw Please have a look at this. There are a few more issues with this file. I'd love to fix them. Just need somehelp. Thanks.

aasfaw:
Fixed by PR #101

PR #70, opened 2019-11-14
Erratum linear_algebra.ipynb Line 139

justcadams:
a start
changed to
state.
Contextually quantum gates apply a matrix transformation to state vectors, which changes the orientation of the q-bits within the quantum computer.


frankharkins:
Thank you! This has been solved in another PR.

PR #71, opened 2019-11-14
Erratum linear_algebra.ipynb Line 163

justcadams:
maatrix 
changed to
matrix.
Contextually Hermitian matrices are equivalent to their conjugate pair.


mrthankyou:
@justcadams 

It looks like that `ch-prerequisites/linear_algebra.ipynb` has a conflicting file change. Can you resolve the conflict? Alternatively, I can take if off your hands and submit a new PR with the fix you suggested here. Thanks.

frankharkins:
Thank you vey much for your contribution, this has been fixed in another PR.

PR #83, opened 2019-11-30
Added full explanation of 2 Qubit case for Grover's algorithm

joschaef:
- Small modification to introduction (clarification on geometrical relationship between |s>, |s'> and |w>)
- Added full explanation + implementation for case of 2 Qubits
- Renamed headers for 3 Qubit Example
- Added images for updated Grover jupyter notebook


aasfaw:
This is great @joschaef -- nice 2-qubit example! Can you rerun all your cells? It looks like the hardware run for your 2-qubit example isn't finished. If you need a hand, let me know :)

joschaef:
OK - I removed some redundant code (which was also present in the 3 qubit case), restarted kernel and ran all.

PR #91, opened 2019-12-05
Fix LaTeX display of composite gate matrix.

dlyongemallo:
The code `gate_latex[0:-4]` is presumably intended to remove the escaped double slashes `\\\\` at the end of the last row of the matrix, but only 2 characters need to be removed because the escape sequences are interpreted before the removal happens. The existing code also erroneously removes the final `0j` which should be displayed.

(It's very confusing because if you look at the raw notebook source code, each slash is escaped again resulting in a sequence of eight slashes `\\\\\\\\`!)


aasfaw:
Very nice catch @dlyongemallo -- thanks!

Issue #99, opened 2019-12-09
The chapter on Quantum Phase Estimation has serious errors


jahanclaes:
The chapter on Quantum Phase estimation has some structural and factual errors. The convention for the QFT circuit that was introduced in the previous chapter was that an input with qubits x1,x2,x3 represented a binary number x1x2x3, and an output y1,y2,y3 represented a binary number y3y2y1, In the chapter on phase estimation, however, the phase estimation circuit is drawn in a way that requires an inverse QFT circuit with the opposite convention to actually work. The examples in the text are run using the wrong convention, which is why there are spurious amplitudes on the basis elements that do not represent the actual phase being estimated. This gives a misleading impression of the algorithm: The algorithm is NOT probabilistic if the phase being estimated can be represented exactly by the ancilla qubits and the gates are applied without errors, but the example makes it seem like even in those circumstances one can still get spurious amplitudes on the wrong basis states.

A simple fix would be to flip the phase estimation circuit to be compatible with the convention in the previous chapter and then rerun the code to generate the correct histograms.

stanescuUW:
@jahanclaes @aasfaw I completely agree with @jahanclaes on the errors in the phase estimation notebook. I have run similar tests in different other environments (PyQuil and Cirq) to confirm that the simulator will **always** return 100% results with the desired bit string for the phase if the phase can be represented exactly in binary. The error is indeed in the inverse QFT; I have fixed it and redone all the notebook with examples that confirm the problems raised by this issue and show correct results. Will start a pull request soon.

PR #133, opened 2019-12-27
Fixed phase estimation code

stanescuUW:
This addresses issue #99 
The phase estimation notebook had serious errors. Fixed the inverse QFT, redone tests. The original 4-ancilla qubit implementation had to be completely removed, so the change is somewhat more exhaustive than initially estimated. The notebook now contains 2-, 3- and 4-ancilla qubit implementations with their respective runs.


frankharkins:
Thank you for your contribution, I eventually decided to completely review the QPE section and have fixed the code in the process.

PR #136, opened 2020-01-01
fixed typos

jeremiahbuckley:
Fixed what appear to be two small typos in the textbook. First, in ch-states/atoms-computation, changed a phrase that said qubits were numbered left-to-right, to say instead that qubits are numbered right-to-left.

Secondly, in ch-states/states-many-qubits, there was a term $P^{zz}_{01}$ where the subscript "01" seems incorrect and it should read "1" instead.


frankharkins:
Thank you!

PR #154, opened 2020-01-14
Update teleportation.ipynb for running on a real device

HwajungKang:
Changed section on running on real device to highlight fact that it does not work, and returns a job error.


frankharkins:
Thanks very much for your contribution. I believe PR#160 will fix the error and enable running on a real device.

PR #160, opened 2020-01-22
fix issues with quantum teleportation notebook

Riuchando:
with the `least_busy` function, it seems that it can choose quantum machines that don't have enough resources. The example is I was given ibmq_essex which only had 1 qubit despite my circuit needing at least 3.

also with the measurements that I was having this issue when trying to run on a physical machine
https://quantumcomputing.stackexchange.com/questions/9380/qubit-measurement-is-followed-by-instructions-error-code-7006 

most likely related to this issue.
https://github.com/Qiskit/qiskit-textbook/issues/142

the stack exchange suggested putting the measurements at the end which seemed to give comparable results.


frankharkins:
Hi Stephen,
Thanks very much for your work. You seem to have a warning:
`/home/stephen/anaconda3/envs/quantum/lib/python3.7/site-packages/qiskit/providers/models/backendconfiguration.py:337: UserWarning: `dt` and `dtm` now have units of seconds(s) rather than nanoseconds(ns).\n",
      "  warnings.warn('`dt` and `dtm` now have units of seconds(s) rather '\n`
I think we need to find a way to resolve or suppress this before we add your changes.

nonhermitian:
The proposed change does not satisfy the steps given in the text, i.e. there are gates conditionally applied based on a measurement result.  This is not supported on the current hardware.  This was made clear in an earlier PR #154 

Note also that the warning listed above emitted by Qiskit by default when accessing a backend configuration.  So there is no resolution for this save for them removing the warning.

frankharkins:
@nonhermitian Thanks for your reply, maybe we could add a couple of sentences explaining the hardware limitations, and why putting the gates before the measurements will still give the correct results?

I have just run the code and do not seem to be getting a warning. Maybe @Riuchando could try running it again? If the warning persists then I think we should at least remove the user specific part, although I would prefer to remove it altogether.

Riuchando:
thanks for the comments, I decided to comment out the code that will not work, and put strikethroughs through the text for now, if there is something better, let me know.

Riuchando:
It's fine that this pr was closed, but is the issue resolved/ will be handled in a different pr?

frankharkins:
@Riuchando, sorry I did not mention it'd been fixed. I am still getting used to github.

PR #168, opened 2020-01-27
Fix Teleportation Chapter

frankharkins:
Added a paragraph on the deferred measurement principle and adapted the code to perform the measurements at the end of the circuit when running on a real machine. Also removed references to qreg and creg in the code.


aasfaw:
This is looking good. Can you add a references section, citing Mike and Ike as well as a reference for deferred measurement? Then we can merge

PR #175, opened 2020-01-28
Fixed wrong statement about inverting noise matrix

jahanclaes:
The textbook claimed that "To obtain an inverse, we need the colums of the original matrix to be orthogonal to each other." This isn't true, you just need the determinant to be nonzero. Matrices with nonzero determinant are a set of measure zero, so you will never actually get one from any numerical procedure.

On the other hand, it IS true that when the determinant is very small, the inverse is very sensitive to noise. This isn't fixed by using the pseudoinverse, though. You just need to do careful error propagation.

### DIFF
--- a/content/ch-quantum-hardware/measurement-error-mitigation.ipynb
+++ b/content/ch-quantum-hardware/measurement-error-mitigation.ipynb
@@ -262,10 +262,6 @@
     "Unfortunately, this is the exact opposite of what we need. Instead of a way to transform ideal counts data into noisy data, we need a way to transform noisy data into ideal data. In linear algebra, we do this for a matrix $M$ by finding the inverse matrix $M^{-1}$,\n",
     "\n",
     "$$C_{ideal} = M^{-1} C_{noisy}.$$\n",
-    "\n",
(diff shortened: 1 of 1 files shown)


content/ch-quantum-hardware/measurement-error-mitigation.ipynb
```diff
@@ -262,10 +262,6 @@
     "Unfortunately, this is the exact opposite of what we need. Instead of a way to transform ideal counts data into noisy data, we need a way to transform noisy data into ideal data. In linear algebra, we do this for a matrix $M$ by finding the inverse matrix $M^{-1}$,\n",
     "\n",
     "$$C_{ideal} = M^{-1} C_{noisy}.$$\n",
```
schwarmco:
```suggestion
     "$$C_{ideal} = M^{-1} C_{noisy}.$$\n"
```
The build fails, because the notebook is no valid JSON anymore. You'll have to remove the comma here.

frankharkins: This change does fix the notebook, @jahanclaes please commit this change before I can merge it

frankharkins:
Thank you!

PR #187, opened 2020-01-31
VQT (Work in Progress)

Lucaman99:
_Discalimer: This pull request is a **work in progress**, more updates have to be made before it should be accepted_

This PR adds a tutorial about simulating the Variational Quantum Thermalizer in Qiskit.

**Changes That Need to Be Made**

- Code blocks need to be broken up more and include more comments about exactly what is going on in each function.
- Sampling simulation needs to be fixed
- Incorporating the sigmoid activation function into the latent space parametrization

If anyone notices anything else that should be changed, other than the things that have already been outlined please leave a comment!


nonhermitian:
What is the difference between this and #403 ?  Can this be closed?

Lucaman99:
@nonhermitian It's not the same thing as #403, but this Notebook isn't really working as I would like it yet, so you can close it for now, and if I get it to work I'll re-open it 🙂

PR #200, opened 2020-02-09
Add default meas pulse

zachschoenfeld33:
## What Changes Were Made?
Default measurement pulse included, pulled from instruction schedule map. Advantage is that this pulse is calibrated frequently, whereas the hardcoded measurement pulse can quickly fall out of date. 

Also fixed a minor issue in how the plots were displayed in Jupyter notebook (remove `%config InlineBackend.figure_format = 'svg' # Makes the images look nice` and instead only use `%matplotlib inline`.

## How Was This Tested?
Ran through full notebook and verified plots generated correctly and with reasonable frequency, Rabi, T1, T2, etc results.


aasfaw:
This is nice. Thanks for updating, @zachschoenfeld33!

PR #210, opened 2020-02-19
no padding, when code is used in a pre (non-inline code blocks)

### LINKED ISSUE #207: Misaligned lines in 1.2
In the [website rendering of 1.2](https://qiskit.org/textbook/ch-states/atoms-computation.html) we see lines that look like

Screenshot 2020-02-19 at 09 59 41

Here the 3 should be directly above the 4, etc, but there is effectively an additional space on the top line.

schwarmco:

Fixes #207 

There is a 4px-padding on `<code>`-tags, which is perfectly fine, when used inline, like here:

image

But fails badly, when used in a multiline block. Because then, the first line (and only the first one) gets intended by 4px, like here:

image

## What Changes Were Made?

I fixed it by applying "padding: 0px;" to `<code>`-tags that are direct childs of `<pre>`-tags, which seems to be always the case, when using multilines.

## How Was This Tested?

I checked all occurances (inline and multiline) and it looks good to me. 

Another approach i tried was to remove the padding completely, but it gets a too little tight on inline-tags then.

### DIFF
--- a/_sass/components/_components.textbook__page.scss
+++ b/_sass/components/_components.textbook__page.scss
@@ -115,6 +115,10 @@ code {
   @include inuit-font-size(14px);
 }
 
+pre > code {
+    padding: 0px;
+}
+
 pre.highlight code {
   display: block;
 }


aasfaw:
Thanks @schwarmco -- this is great

PR #221, opened 2020-02-24
Qutrit chapter: remove anharmonicity derivation

zachschoenfeld33:
## What Changes Were Made?
Removed line on deriving the anharmonicity from the Hamiltonian because the math was not correct. The proper way to do this is via perturbation theory (which isn't too challenging, but I don't think it's that important so I removed these lines completely).

## How Was This Tested?
Nothing changed from prior version, which was tested on `ibmq_armonk` system.


aasfaw:
Yeah shouldn't be too bad to fix, will merge this for now

PR #231, opened 2020-02-28
replaced deprecated `channels_to_plot` to `channels` in calls to pulse.Schedule.draw()

warrenalphonso:
In schedule.py in qiskit-terra, the `channels_to_plot` parameter for the draw() function has been deprecated in favor of `channels`.

## What Changes Were Made?
Replaced all `channels_to_plot` with `channels`

## How Was This Tested?
Nothing was functionally changed so I don't think tests are necessary; I wasn't sure if rerunning the cells was okay since it would change some of the matplotlib graphics with slightly different data. I got the same output when running my own code in a different repo when reading through the textbook.


frankharkins:
thank you!

PR #248, opened 2020-03-08
updated 3.6

starktech23:

Typos in chapter 3.6 for QFT

Fixes #your_issue_number_here

## What Changes Were Made?

1. "Let’s now encode ~~the~~ a number in the computational basis. We can see the number 5 in binary is 101"
2. Ch 3.6 Section 6 Steps 3,4,5 Effect of CROT1 gate to |x1> depending on |x3> = (exp(2πi/2^3)x3)|1>

## How Was This Tested?


starktech23:
Quick update: I think the Python version in this PR needs to be edited back to 3.7.6, I was probably using an older version.

PR #261, opened 2020-03-12
eigenvector and eigenvalue using same letter

fguevaravas:
fix for typo: using $v$ for both eigenvector and eigenvalue was confusing in the explanation of eigenvalues of the exponential of a matrix are the exponentials of eigenvalues of the matrix


frankharkins:
I am closing this as it will be incorporated in #612

PR #262, opened 2020-03-12
Add CTA analytics

delapuente:
I've added a `trackCta()` function to the global object `window`. It takes an action parameter, which is a string describing the call to action. @frankharkins feel free to suggest me other descriptions.

With these changes, we can keep track about enabling interactivity, going to IQX and running some specific cell.


frankharkins:
Awesome! Thanks Salva

PR #263, opened 2020-03-13
Refactor the code for creating use-hardware cells.

delapuente:
Includes re-styling of "the use hardware notice".

Add analytics to the "run using IQX" links.

### DIFF
--- a/_includes/js/thebelab.html
+++ b/_includes/js/thebelab.html
@@ -62,6 +62,12 @@
             const hardwareCells = document.querySelectorAll(hardwareQuery)
             hardwareCells.forEach((cell) => {
                 cell.prepend(usesHardwareNotice.content.cloneNode(true))
+                cell.querySelector('a').addEventListener('click', (evt) => {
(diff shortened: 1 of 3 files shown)


_includes/js/thebelab.html
```diff
@@ -25,42 +24,45 @@
...
-		    hardwareWarning += '</a></code>'
-
-		    hardwareCell.innerHTML = hardwareWarning + hardwareCell.innerHTML
-	    });
+
+
+            // Create link to IQX and hardware notice template
+            const currentPath = '{{ page.url }}'
+            const notebookPath = currentPath.replace(/\.html$/, '.ipynb')
+            const iqxUrl = `{{ site.iqx_url }}/content/${notebookPath}`
```
frankharkins: This link is not directing me to the right place. I had the same problem: For some reason `page.url` always points to the first page you visited, and not the notebook page you're on. This is why I ended up using `window.location.pathname`.

PR #267, opened 2020-03-18
Modified hello_quantum.py to make it compatible with the textbook

delapuente:
Please notice I did not add it to the TOC.


quantumjim:
👍

PR #285, opened 2020-03-26
Add missing qiskit_textbook/__init__.py

### LINKED ISSUE #284: Unable to import qiskit_textbook from jupyter notebook
## Describe the Issue

With a recent change [1] I'm unable to import `qiskit_textbook` in a jupyter notebook. It appears to be because there is no `__init__.py` under that package.

We're installing by doing `pip install -r requirements.txt` from the `qiskit-textbook-src` directory.

[1] https://github.com/Qiskit/qiskit-textbook/commit/b81f279d23b0eb1457d475fec68e7cdf0af55654#r38077999

## Expected Solution

The refactor of the textbook package should be pip installable such that the `qiskit_textbook` package can be imported.

The weird thing is if I git clone this repo and do the same `pip install -r requirements.txt` then I'm able to start a python session and import the `qiskit_textbook` package so I'm not sure why that's different, but python packages should have an `__init__.py` in them regardless.

mriedem:

Fixes #284 

## What Changes Were Made?

This is pretty self-explanatory. #284 has more background details.

## How Was This Tested?

This makes installing and importing `qiskit_textbook` work again in jupyter notebooks.

Nope.


frankharkins:
Thank you!

PR #304, opened 2020-04-01
4.1.1hhl 2B step 5 added forgotten C in normalization

HTransistor:
In step 5 calculating the normalization constant the C^2 was missing in the denominator.
4 1 1hhl_2B_step5

In the example3 when writing |b> in the eigenbasis the sum goes from 0 to N-1,
but the eigenstates were called u1 and u2.
4 1 1hhl_3example_b_in_eigenbasis

Fixes #your_issue_number_here

## What Changes Were Made?
1. C^2 added as factor in the denominator.
2. Corrected the sum going from j=1 to 2 (as these are the eigenstates in this example)

## How Was This Tested?


frankharkins:
Thanks for raising this issue. I believe the constant C does not need to be carried forward as the equation in 5 is correctly normalised. This is because the vectors $|u_j\rangle$ are an orthonormal basis ([see the second-to-last property here](https://en.wikipedia.org/wiki/Hermitian_matrix#Properties)).

I agree with point 2 though.

HTransistor:
To point 1:
Step 4 is normalized, yes, but when you now measure a 1 you get the right part of the bracket, so: $\sum_{j=0}{N -1} b_j * C / \lambda |0\rangle |u_j\rangle$.
If you want to normalize it now there is a C^2 missing in the sqrt of the denominator.

frankharkins:
I'm not quite sure I follow. The state of the register needs to be normalised at all times to ensure the probabilities add up to 1. Unless I'm mistaken, adding the C^2 to step 5 means the state is not normalised.

HTransistor:
Mentioning that it is normalized too, would be good.
I think the problem is the measurement.
After measurement (1):
$\sum_{j=0}{N -1} b_j * C / \lambda |0\rangle |u_j\rangle$.
Norm (2):
$\sqrt( \sum_{j=0}{N -1} (b_j * C / \lambda )^2 )$
As C is constant, take it out the bracket and sqrt(C^2) = C, so (3):
$C * \sqrt( \sum_{j=0}{N -1} (b_j / \lambda )^2 )$

Now devide (1) by (3) to normalize:
Result I was wrong. The C cancels :D

frankharkins:
Awesome! Thanks so much for working through this with us, glad we found a resolution.

PR #348, opened 2020-04-25
Corrected definition of Pauli-Y matrix

paniash:
The mathematical definition of Pauli-Ymatrix (in terms of outer product) was incorrect, i.e. the signs were flipped.

## What Changes Were Made?
Changed the signs of the relevant terms.
## How Was This Tested?
With a pen and a piece of paper.


johnzl-777:
Also ran into the same issue, but verified with `numpy` matrices representing raw bras and kets.

Would love to see this corrected!

PR #367, opened 2020-05-07
Fix #360

### LINKED ISSUE #360: 1st entry in After CNOT-gate is wrong.
## Describe the Issue
 1st entry in After CNOT-gate is wrong.
## Expected Solution
It should be |00> + |01>
superdense_table2

aditya-giri:
This fixes #360 
Since the table was being displayed as an image as opposed to markdown, I had to update the image itself. The dimensions have changed a little bit but the overall quality of the image is about the same.
The diff can be viewed using the `rich diff` option under `Files changed`.


frankharkins:
wonderful, thank you!

Issue #395, opened 2020-05-24
Sec 1.3 Updates


nonhermitian:
Sec 1.1 needs to be worked on.

> This is fine, as a classical bit always has a definite state. With qubits however, this restriction is lifted and we have probabilities of measuring different states.

This reads oddly.  A qubit has a definite state as well.  

> Each state a qubit can be measured in has an amplitude (a complex number), and we need to keep track of these complex numbers.

Why complex?

>  have non-zero amplitudes, the qubit is said to be in a superposition of the states |0⟩ and |1⟩.

Superposition is just a fancy way of saying linear combination.  This should be explained to the reader.

> %config InlineBackend.figure_format = 'svg' # Makes the images look nice

This is really only needed for the textbook to generate svg files upon building.  This should be hidden from the reader as it has no learning purpose.

> initial_state = [1/sqrt(2), complex(0,1/sqrt(2))] 

This can be simplified to `initial_state = [1/sqrt(2), 1j/sqrt(2))] `

For Sec 2.1.  |<psi|x>|^2 should be reworked as |<x|psi>|^2. This is because expressing the desired state psi in the direction of x is written |x><x|psi>.  So it is a bit clearer to reorder the terms for clarity.

> The second implication is that the total phase of the qubit (global phase) does not matter to us, only the difference in phase between |0⟩ and |1⟩ (relative phase). This is because the global phase disappears when we calculate the measurement probability.

This needs to be reworked.  Even without making a measurement the global phase does not matter.

> And find it in the state |0⟩, if we measure again, there is a 100% chance of finding the qubit in the state |0⟩.

Immediately after the measurement.

> In fact, measuring not only changes the state of our qubits, it destroys the superposition of our qubit, replacing it with one of two definite states. As a result, we almost always place the measurements at the end of our circuit.

I do not recall where it was previously mentioned that superposition is important, so unclear as to why I do not want to destroy it.

> Instead of having 𝛼 and 𝛽 be complex, we can confine them to the real numbers and add a term to tell us the relative phase between them

This reads oddly.  Namely you just said alpha and beta are complex, but now saying not really, they are real and a phase comes out of nowhere.  Better to say that we pull out a phase factor such that the |0> term becomes real (by convention), resulting in the additional phase factor in front of beta.

> (Qiskit has a function to plot a bloch sphere, plot_bloch_vector(), but at the time of writing it only takes cartesian coordinates. We have included a function that does the conversion automatically).

[...]

PR #407, opened 2020-05-26
Improve 1.3

quantumjim:
Some improvements to part 1 so far. This restored some good parts of sections that have been removed, and addressed some of the comments in issue #395 .

To do:
* address the other comments
* maybe remove the Bloch sphere?


nonhermitian:
> If you are not familiar with vectors, you can essentially just think of them as lists of numbers which we manipulate using certain rules. 

We have linear algebra as a prerequisite and go over vectors in that section, so should assume the reader knows this information.

> This is a lot of notation to take in all at once. First let's unpack the weird $|$ and $\rangle$ 

This is not weird, but may be unfamiliar to the reader.  However, again, linear algebra is a prerequisite, and this notation is used there.

> If you are not familiar with vectors, you can essentially just think of them as lists of numbers which we manipulate using certain rules. 

Same.

> For example, thebvelocity

Spelling

> If you are familiar with vectors from your high school physics classes, you'll know that these rules make vectors well-suited for describing quantities with a magnitude and a direction. For example, thebvelocity of an object is described perfectly with a vector. However, the way we use vectors for quantum states is slightly different to this. 

The vectors in QM also have amplitudes and directions.

This chapter highlights a general disconnect between the linear algebra "prerequisite" and the text.  A prerequisite is something that is necessary to have ahead of time.  Thus we should either assume that the reader knows it, or drop the notion of linear algebra as a prerequisite, and move it to an appendix.

quantumjim:
Thanks for the comments. But regarding the 'linear algebra as a prerequisite' thing, I don't think reading a wall of text on linear algebra should be a prerequisite. It is better for us to introduce as we go along, and refer to the lin alg section as a resource that can be used for people to brush up.

quantumjim:
There is a lot that people can get out of the textbook without needing to stare at the Maclaurin series, etc. What's the point in trying to start the textbook off in a gentle and ped

nonhermitian:
I would agree.  The lin alg section is quite overwhelming.  But it does mean a decision about readers knowledge level needs to be solidified.  I see you are doing this in #409

quantumjim:
Since the improving the Bloch sphere part is dependent on a Terra PR (for alternative measurements). We can do that in a future PR. I've taken the WIP off for now.

I haven't merged myself, because I don't like overusing my superpowers.

frankharkins:
This is much better, thanks @quantumjim.

PR #416, opened 2020-05-31
A few changes to QFT, QPE, Simon's algo and Grover's algo nbs

### LINKED ISSUE #393: QFT chapter uses CROT instead of UROT
## Describe the Issue

Given that in the QFT chapter the UROT gate is defined as a single qubit gate, and the CROT gate is defined as a conditional UROT gate, the circuits (and some of the text) in the QFT chapter are incorrect.

For example, in the image below, the gates should really be UROT and not CROT, the same way a CX gate is represented by a dot on the control qubit and an X gate (and not a CX gate) on the target.
image

I also believe that text like:
```After the CROT2 gate on qubit 1 controlled by qubit 2, the state is transformed to …```
should ideally be written:
```After the UROT2 gate on qubit 1 controlled by qubit 2, the state is transformed to …```
but I understand that this is debatable and can be understood either way.

EDIT: Also, the last qubit seems to be missing a H gate.

PS: I'll pick this up :)

aditya-giri:
Fixes #393 and a couple of smaller issues in other notebooks

I changed the numbering in section 1.2 of QPE from `1,2,3,...` to `i, ii, iii,...` because the arabic numerals weren't being displayed correctly on the final page (refer image below).
Roman numbers seemed to work better.
qpe-numbering


frankharkins:
This is wonderful, thank you!

PR #417, opened 2020-06-01
Rename Eve Charlie in Superdense Coding

ikkoham:

Eve is an eavesdropper most of the time in this context, so Charlie is better here.

## What Changes Were Made?

Rename Eve Charlie.

I also fix Eve's message to Alice's message because message sender is Alice.
see: https://github.com/Qiskit/qiskit-textbook/compare/master...ikkoham:rename-eve-charlie?expand=1#diff-399841aaf8bdfb6d2b5cd0071ba898c1R111


frankharkins:
Thank you!

PR #424, opened 2020-06-03
Ensure the engine script is initialized after the engine is loaded

### LINKED ISSUE #325: Search does not work
## Describe the Issue
Searching "circuit" does nothing.

## Expected Solution
Search works

delapuente:
Fixes #325

## What Changes Were Made?

Factor out a helper function to asynchronously load scripts and use it to coordinate the initialization of some libraries such as lunr, the search engine.

## How Was This Tested?

Compared errors in the console before and after the modification.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -83,7 +83,6 @@
   {% include js/interact-update.html %}
 
   <!-- Lunr search code - will only be executed on the /search page -->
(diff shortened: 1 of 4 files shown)


frankharkins:
Awesome stuff, thanks Salva.

PR #431, opened 2020-06-04
Add new section 'The Case for Quantum Computers'

### LINKED ISSUE #386: Need to better emphasise motivations for quantum computing
## Describe the Issue

There is surprisingly little explaining the motivations of quantum computing early on in the textbook.

## Expected Solution

I think something that should be covered first is big-O-notation and why different algorithm run times are so important. I would then follow this with a short list of applications which are limited in classical computing but have potential to be solved with QC.

quantumjim:
To address the need to justify why quantum computers will be needed. Closes #386. Intended as to be section 1.5

Starts with a callback to 1.2, to reflect on the complexity of addition. This leads into big-O notation, complexity theory, and the theoretical notion of a digital computer.

Some justification for quantum computers is then done with an analogy I like to use, which is that wave-particle duality becomes digital-analogue duality when we think about computation. It gives a taste of why quantum computation is unique. The rest of the textbook's job is to establish that concretely.

It will end with some hand waving discussions of when a speedup can be expected and how QCs will be the best thing since sliced bread. But I don't have my propaganda hat on today.


quantumjim:
Even though this is draft, it is still ready to review.

cognitivecomp:
Hi @quantumjim , thanks for pulling this together! I agree with @frankharkins 's points. Here are a few more comments / ideas:

- Typo: "where don't need to carry a 1"
- Typo: "require to expressed the same number"
- In the discussion on complexity theory, might it be worth referencing BQP / how we assume BQP relates to other complexity classes? If so, a chart could help, see e.g. the chart on P. 67 (6th page from the start) here: https://www.cs.virginia.edu/~robins/The_Limits_of_Quantum_Computers.pdf
- In Section 4, you mentioned "We will see that quantum computers can solve problems with a radically different complexity to digital computers." We could perhaps expand this with a comment along the lines of "It is the only known technology that can be exponentially faster than classical computers for certain tasks, potentially reducing calculation times from years to minutes."
- For the examples, we could potentially add a few words about the kinds of problems for which quantum computing is particularly promising (no need to load a large classical dataset, inherent scaling limits of the best classical algorithms). As @frankharkins mentioned, we could then expand the list of examples and include e.g. machine learning, materials discovery, option pricing, protein folding, supply chain optimization, ... If people want to read on, we could link some of the quantum industry Expert Insights such as https://www.ibm.com/downloads/cas/EVBKAZGJ .

Feel free to delegate if there's anything you'd like me to incorporate.

quantumjim:
The new commit was me just focusing on finishing the unfinished bit. So now it's more complete, but all typos are still there.

I agree the comments so far, except that disucussion of complexity classes might be more trouble than its worth. We'll only find out if we try!

Since two people have volunteered for delegated tasks, might I suggest that you both take care of implementing your suggestions and finishing this up? Then I can get on with creating my next wall of text!

Issue #448, opened 2020-06-15
Index required


quantumjim:
## Describe the Issue

Textbooks usually include an index, which allows the reader to determine where in the book they can find certain topics. Just searching the text is not a good substitute for this.

## Expected Solution

Add keywords to each section, and use this to automatically create an index.

Vismai-Khanderao:
I feel like this may be a good addition to the textbook backlog project

frankharkins:
Thanks for your contribution. I'm closing this issue as this repository is being archived. You can find the latest version of the Qiskit Textbook online at <https://qiskit.org/learn/>, and the new content repository at [Qiskit/textbook](https://github.com/Qiskit/textbook). Please feel free to re-open this issue there if it's still relevant.

PR #449, opened 2020-06-15
add indexer

quantumjim:
Fixes issue #448 . 

This looks through all notebooks to find a dictionary called 'keywords', and then uses the content of these to make an index. This is written to a file 'index.md' at the top level.

For example, if the following was added to 'atoms-computation.ipynb'

`
    keywords = {'Topics': ['Gates', 'Boolean logic gates', 'AND gate', 'XOR gate'], 'Commands': ['`x`', '`cx`', '`ccx`']}
`

The 'index.md' file would be

### `ccx`
* [atoms-computation](content/ch-states/atoms-computation.ipynb)

### `cx`
* [atoms-computation](content/ch-states/atoms-computation.ipynb)

### `x`
* [atoms-computation](content/ch-states/atoms-computation.ipynb)

### AND gate
* [atoms-computation](content/ch-states/atoms-computation.ipynb)

### Boolean logic gates
* [atoms-computation](content/ch-states/atoms-computation.ipynb)

### Gates
* [atoms-computation](content/ch-states/atoms-computation.ipynb)

### XOR gate
* [atoms-computation](content/ch-states/atoms-computation.ipynb)


quantumjim:
This is something I made for the community tutorial, but it never really got used. Perhaps it will be useful here. It was motivated session with educators last year, where they expressed the desire for an index for educational materials.

It could probably do with some tweaking here and there. For example, do we really want the 'index'.md file to have filenames in?

frankharkins:
This is a great idea, although the links have reminded me of a related problem that needs a solution: I don't believe the `.ipynb` extensions are converted to `.html` on the website. I will try to find out if this is the case and fix it if possible before we merge this.

aasfaw:
Super useful feature, let's wait for Frank's extension check, then we can merge

PR #451, opened 2020-06-15
Q.Counting: Explained N-M, tried to clarify comments on precision.

frankharkins:
I explained why the code outputs N-M instead of M, and tried to clarfy some of the comments about precision. The comments on precision are still WIP and need some other opinions, I am alternatively considering removing them altogether.

The explanation of N-M references the changes added to the Grover chapter in [this PR](https://github.com/tbcdebug/qiskit-textbook/pull/1).


HwajungKang:
I am not sure about this PR... Do you want to be approved for the update about the explanation for the code output N-M instead of M regardless of the WIP status for the second issue for now or you want to be approved after you resolve the other as well?

frankharkins:
@HwajungKang please only approve once you're happy with all the changes in the PR. The changed line (about precision) is currently:

> And we can see we have (approximately) the correct answer! This is not _that_ precise as we are only using 4 qubits. You might think that 4 qubits would be enough to give us an exact answer, but remember that there are two register measurements that will give us the same answer ((2^t) * theta}/{2*pi}). This halves the number of unique answers and loses us one bit of precision. We then lose some precision to the squared sine function.

HwajungKang:
>there are two register measurements that will give us the same answer ((2^t) * theta}/{2*pi})

Shouldn't it be more like " there are two possible measurement outcomes that give us the same answer for the number of solutions" ?

> This halves the number of unique answers 

you mean there is only one unique answer instead of two for the number of solutions? If that is what you mean, you already stated it.

> and loses us one bit of precision.

First of all, I guess that you mean accuracy not precision?
Also, this part, as I told you the other day, I am not so sure if this is a correct statement...
maybe, I am wrong, but I don't know if the number of different eigenvalues being measured has anything to do with the accuracy of the estimated value for theta since the accuracy of the estimation would be determined by the number of counting bits.

frankharkins:
With 4 counting qubits, the reader might expect an exact answer as there are enough qubits to hold M exactly. I was trying to make the point that, from 16 possible register measurements, we can only get 8 values for M. 

After seeing the confusion I now think it may be better to remove this. I don't want to give the impression that having a register in superposition reduces the accuracy of a measurement, and the squared sine function is enough to explain any loss of accuracy.

PR #465, opened 2020-06-25
Removed redundancy

paniash:
Removed redundancy in code.

## What Changes Were Made?
Changed `display(qc.draw())` to `qc.draw()` wherever applicable.

## How Was This Tested?
IPython shell.

### DIFF
--- a/content/ch-gates/phase-kickback.ipynb
+++ b/content/ch-gates/phase-kickback.ipynb
@@ -8199,7 +8199,7 @@
     "qc.h(0)\n",
     "qc.h(1)\n",
     "qc.cx(0,1)\n",
-    "display(qc.draw())\n",
+    "qc.draw()\n",
     "\n",
     "# Let's see the result\n",
     "statevector_backend = Aer.get_backend('statevector_simulator')\n",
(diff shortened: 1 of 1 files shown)


frankharkins:
@paniash this doesn't work in Jupyter Notebooks (the recommended course of reading the textbook). Since the notebook only displays `qc.draw()` automatically when it is the last line in a cell. Maybe we could add a comment explaining this for other users instead of removing them?

paniash:
@frankharkins Please correct me if I'm wrong but doesn't `%matplotlib inline` enable `qc.draw()` to work on Jupyter Notebooks?
If not, then please allow me to follow your suggestions and make another pull request.

paniash:
I tried the code block in Jupyter Notebook and it seems to work fine on my system.

frankharkins:
This is what I get (even when including `%matplotlib inline`). The circuit does not draw without `display()`. 

Screenshot 2020-06-25 at 15 17 38

Do you know if there's a way to fix this on my device? If so we should add it to the readme.

paniash:
I see what the problem was. When running just `qc.draw()`, I get a separate window displaying the circuit (it doesn't draw in the notebook itself). Only when running `display(qc.draw())`, do i get a separate window as well as the circuit on the notebook.

I think it will be better to include this as a comment. Also, shouldn't this change be propagated to the previous chapters as well since those contain only `qc.draw()` instead of `display(qc.draw())`. I had been using IPython all this while and as unaware of this difference.

nonhermitian:
The circuit will not draw because only the last fig is captured. Matplotlib inline is actually legacy code. It is done by default these days. The display is forcing the output. One could also print the circuit, although may not be ideal.

PR #471, opened 2020-06-26
Fixing a typo in measurement-error-mitigation

### LINKED ISSUE #464: Significant typo in 5.2 Measurement Error Mitigation
## Describe the Issue  

About 1/3 of the way down there is a paragraph saying:
> The information gathered from the basis states  |00⟩ ,  |01⟩ ,  |10⟩  and  |11⟩  can then be used to define a matrix, which rotates from an ideal set of counts to one affected by measurement noise. This is done by simply taking the counts dictionary for  |00⟩ , normalizing it it so that all elements sum to one, and then using it as the first column of the matrix. The next column is similarly defined by the counts dictionary **obtained for  |00⟩ , and so on.**

## Expected Solution
The bold text should read
**obtained for  |01⟩ , and so on.**

rochisha0:
Significant typo in 5.2 Measurement Error Mitigation #464
Fixes #464

## What Changes Were Made?

|00> to |01> in the line -
The information gathered from the basis states  ||00⟩ ,  ||01⟩ ,  ||10⟩  and  ||11⟩  can then be used to define a matrix, which rotates from an ideal set of counts to one affected by measurement noise. This is done by simply taking the counts dictionary for  ||00⟩ , normalizing it so that all elements sum to one, and then using it as the first column of the matrix. The next column is similarly defined by the counts dictionary obtained for  **||01⟩** , and so on.

### DIFF
--- a/content/ch-quantum-hardware/measurement-error-mitigation.ipynb
+++ b/content/ch-quantum-hardware/measurement-error-mitigation.ipynb
@@ -169,7 +169,7 @@
     "\n",
     "Here the first element is that for `'00'`, the next is that for `'01'`, and so on.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


rochisha0:
Is this PR merged? It is my first contribution actually!

PR #476, opened 2020-06-29
Fixed links pointing to phase estimation and Fourier transform sections

diemilio:
Links to the phase estimation and Fourier transform sections were pointing to their respective python notebooks so they were not working on the textbook website.

## What Changes Were Made?
Replaced the links so the point to the right html sections.

### DIFF
--- a/content/ch-algorithms/shor.ipynb
+++ b/content/ch-algorithms/shor.ipynb
@@ -1245,7 +1245,7 @@
    "source": [
     "## 2. The Solution\n",
     "\n",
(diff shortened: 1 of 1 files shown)


frankharkins:
Thank you for this. I actually think the links should be converted as part of the build process. This is so people can use the links to move between notebooks when using the textbook locally or on IQX. I am looking into a fix.

diemilio:
Thanks Frank; understood. I will close the PR then.

PR #499, opened 2020-07-08
Fix maxcut computation from samples in qaoa notebook

### LINKED ISSUE #500: [4.1.3] QAOA notebook has wrong cost computation function
## Describe the Issue

In the following notebook: https://qiskit.org/textbook/ch-applications/qaoa.html `cost_function_C` function computes the MaxCut cost for a sample. The sample is created in the following line, [link to the notebook section](https://qiskit.org/textbook/ch-applications/qaoa.html#Evaluate-the-date-from-the-simulation):
```
    x         = [int(num) for num in list(sample)]
```

The samples are in little endian format, i.e. applying `x[0]` on a 4-qubit circuit will results in a sample string `0001`. The qubits, however are stored in reversed order

## Expected Solution

Reverse the label of the sample:
#499 

The graph used in the example is symmetric with respect to relabeling of qubits `0...n -> n..0`. Maybe the issue would be spotted if another graph was used. I used a square with one diagonal, adjacency matrix:
```
[[0. 1. 0. 1.]
 [1. 0. 1. 1.]
 [0. 1. 0. 1.]
 [1. 1. 1. 0.]]
```

danlkv:

Fixes #500 
## What Changes Were Made?

Use reversed key of samples dictionary, to account for littlie-endianness of resulting states

## How Was This Tested?

Tested on a graph with following adjacency matrix:
```
[[0. 1. 0. 1.]
 [1. 0. 1. 1.]
 [0. 1. 0. 1.]
 [1. 1. 1. 0.]]
```

### DIFF
--- a/content/ch-applications/qaoa.ipynb
+++ b/content/ch-applications/qaoa.ipynb
@@ -3,6 +3,16 @@
   {
    "cell_type": "markdown",
    "metadata": {
+    "toc": true
+   },
+   "source": [
+    "<h1>Table of Contents<span class=\"tocSkip\"></span></h1>\n",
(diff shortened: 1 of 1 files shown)


DaniloZZZ:
Just spotted that my table-of-contents plugin for jupyter notebook added a table of contents, fixed in a second commit

PR #514, opened 2020-07-10
Adds Section on The Density Matrix (Work in Progress)

diemilio:
Adds section on the density matrix and mixed states (proposed in #8 ). This is a **_work in progress_**

Ideally, this will be part of a chapter in Quantum Information (QI), so the notebook was placed under a new folder (ch-quantum-information). However, it could also be placed at the end of chapter 2 or at the beginning of chapter 3 while more content is added to the QI chapter.

Please add a comment if you find any typos, errors, or technical inconsistencies.


nonhermitian:
I think what is missing here is the "why".  Why are you introducing the density matrix here?  Why is it important to know in the context of quantum computing?  Also nothing about entanglement is discussed even though the machinery to handle this is being introduced.

diemilio:
Hi @nonhermitian, thank you so much for your feedback. 

Regarding: "I think what is missing here is the "why". Why are you introducing the density matrix here?". From the discussion in #8, I believe the idea will be to have a new chapter on "Quantum Information", for which a section on the density matrix and mixed states will be necessary. Would adding a paragraph at the beginning describing how mixed states are used represent, for example, the effect noisy communication channels, noisy quantum hardware, and the environment have on pure states?

Regarding: "nothing about entanglement is discussed even though the machinery to handle this is being introduced." I apologize. I tried to make this section as self-contained as possible and added links to concepts already introduced in other sections of the textbook (like statevectors, outer products, the Bloch sphere, etc.). It seems I forgot to add one when I first introduced entangled states. Will adding a link to this [section](https://qiskit.org/textbook/ch-gates/multiple-qubits-entangled-states.html#entangled) suffice? 

Thanks again!

nonhermitian:
So you need to think of this from the readers point of view: Why should I be reading this?  You need to motivate why understanding the density matrix formalism is worth the effort.  Indeed, the vast majority of the textbook does not need a density matrix to understand.  Also the level of mathematical rigor is much higher here than in the other sections.  Indeed, many places do not expect users to be familiar with linear algebra.  Thus this chapter needs to be interpreted in the context of the level of the rest of the textbook.  I would also explicitly state when the density matrix should be used verses just the statevector.  Many people are confused about this.

Also the entanglement part actually needs to be formally expressed here.  The section you want to link to says nothing about entanglement, just that it exists and Bell states are entangled.  Again, this is because the textbook assumes very minimal knowledge of mathematics.  So if going down the route of a full density matrix chapter, entanglement needs to be discussed.  At minimum I would discuss von-Nuemann entropy and purity.  Explicit examples for bi-partite systems should also be included.  One can also measure these values experimentally, so having an example of extracting the purity of one of the qubits in a Bell state, for example, should also be given.

diemilio:
Thanks again @nonhermitian, this is great feedback!

> "… you need to think of this from the readers point of view: Why should I be reading this?"
Understood. I discussed with Frank and he gave me some pointers on how to improve this. I will work on it.

>  “… the level of mathematical rigor is much higher here than in the other sections."
Can you please let me know if there is something specific you are referring to? I tried to avoid using any concepts that hadn't been previously introduced in the book. Parts 1, 2 and 5 only require knowing about _state vectors_, _outer products_, and _Pauli matrices_, all of which are covered in chapters 1 and 2 of the textbook and are properly referenced. Parts 3 and 4 are a bit more dense, but they really only require prior knowledge on _unitary matrices_ (covered in the _Linear algebra_ section and revisited in the _Proving Universality_ section), the _inner product_, and expressing _states as superpositions_ (also covered in the “Linear algebra” section). Is there something in particular you would like to see changed/clarified?

> “I would also explicitly state when the density matrix should be used verses just the statevector. Many people are confused about this.”
I tried doing this at the very beginning of the section, but I can see how it might be easy to miss. Frank gave some ideas on how to improve it, so I'll work on it.

> "Also the entanglement part actually needs to be formally expressed here... The section you want to link to says nothing about entanglement, just that it exists and Bell states are entangled."
I apologize; I misinterpreted what you meant the first around. I 100% agree with you on this; however, I think this should actually be covered in a separate section dedicated just to entanglement. And I would argue that that section should go after introducing the density matrix because in order to talk about Von Neumann entropy you need to understand the concept of trace and the density matrices in general.
My understanding is that Abe, James and Frank are still working out the details on how the QI chapter will be structured, but I think it would be nice to have a section that introduces Schmidt decomposition, then covers the subject of quantifying entanglement (using both the Schmidt number and Von Neumann entropy), and then talk about entanglement as a resource (maybe even cover the fungibility of entanglement). Now obviously this would have to be done keeping the Qiskit framework in mind, and how these concepts fit within it.

> “Explicit examples for bi-partite systems should also be included. One can also measure these values experimentally, so having an example of extracting the purity of one of the qubits in a Bell state, for example, should also be given.”
This is a great idea. I did briefly touch on this under the “_Trace of the Square_”, but I can definitely rename that subsection to something like “State Purity”, and give some examples. Frank told me he was also going to give this a little bit more thought.

Thanks again for the feedback; I'll stay in touch with Frank on how to proceed on some of these items, and start working on the others.

content/ch-quantum-information/density-matrix.ipynb
```diff
@@ -0,0 +1,1074 @@
...
+   },
+   "source": [
+    "# The Density Matrix & Mixed States"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The Density matrix, or density operator, is an alternative way of expressing quantum states. Unlike the state vector notation introduced in the [Representing Qubit States](../ch-states/representing-qubit-states.ipynb) section, the density matrix provides a convinient and elegant way to introduce the notion of classical statistical distributions to quantum systems. For example, when working with real quantum computers, the state of a qubit is subject to the noise of its enviroment and to uncertainties generated by the qubit's control hardware. This state is therefore not only a function of its own quantum probability amplitudes, but it also depends on the probability distribution of the noise around it. We call these type of states \"mixed\" to distinguish them from the \"pure\" states that don't depend on these externalities. \n",
```
nonhermitian, line 18: > the probability distribution of the noise around it

I am not sure what you mean by this?

content/ch-quantum-information/density-matrix.ipynb
```diff
@@ -0,0 +1,1074 @@
...
+    "\n",
+    "So, the advantage of using the density matrix formalism rather than state vector representation, is that it allows us to use the same mathematical language to express both \"ideal\" quantum states (pure states), as well as the more realistic states we encounter when working with actual quantum hardware (mixed states).\n",
```
nonhermitian, line 20: > the more realistic states we encounter when working with actual quantum hardware (mixed states)

The state of the quantum computer is always a pure state so this reads oddly.

content/ch-quantum-information/density-matrix.ipynb
diemilio: OK. It seems the edits didn't help and this still isn't clear. I will rewrite it.

IvanIsCoding (commented):
Commenting after the reading group organized by @frankharkins:

* First things first, this is great work and a necessary addition to the textbook 😃 
* The Reduced Density Matrix section may be rewritten to have a better flow and be less overwhelming
* The notebook may be split into two because it is fairly long. Part one could go until the Properties of Density Matrix, it is a good conclusion point. Part two could have the additional details with discussion of Reduced Density Matrix + Bloch Sphere.

Ellaborating a bit more on the Reduced Density Matrix point. That section starts by introducing a new linear algebra concept, partial trace, which is overwhelming. It feels like an alien concept, there's little motivation behind why we need partial trace. Nevertheless, the conclusion at the end analysing the Mixed State is great and clarifies things.

Hence: why not rewrite the section to start with the motivating example, and then introduce partial trace and Reduced Density Matrix as tools to analyse it? It makes it an easier read than starting with pure linear algebra definitions.

frankharkins (approved):
Looks good, thanks!

amirebrahimi:
> Looks good, thanks!

When does this show up on the website?

nonhermitian:
This does not use the correct Qiskit qubit ordering in the tensor product.

diemilio:
Thank you @nonhermitian. It has been a little while since I last looked into this, but I will go over it, find where the ordering is incorrect, and add it to my to-do list.

dhruvbhq:
Hi, thanks for this great addition!

- In [5.5.1 Pure States](https://qiskit.org/textbook/ch-quantum-hardware/density-matrix.html#1.-Pure-States-), the content states: "We understand that if we were to perform a measurement of this state, the outcome will be probabilisitc". It should be: "We understand that if we were to perform a measurement of this state, the outcome will be probabilistic", i.e. there's a typo in "probabilistic".
- Also see #1122 (had I known about this discussion place, I wouldn't have created a separate issue - my bad)

dhruvbhq:
Everyone, to be able to track previous and further suggestions to this section, @diemilio has suggested to move this discussion to the separate issue #1122 . Inputs are welcome there!

PR #531, opened 2020-07-17
2.3, 2.4 ja translation

kifumi:
## What Changes Were Made?
Japanese translation of 2.3 Phase Kickback and 2.4 More Circuit Identities.


frankharkins:
Not sure why this is failing, apparently a problem with `make book`, although it seems to work fine on my machine. I will look into it, but will merge for now as it's a problem with Jupyter Book and not the content.

kifumi:
Thank you, Frank.

PR #540, opened 2020-07-18
Text changes in Simon.ipynb

Sahel13:
## What changes were made?

In section '1a. Simon's Problem', an 'f' was missing in 'one-to-one'. Also, in section '1b. Simon's Algorithm' - 'Quantum Solution', a hitherto unexplained variable 's' was introduced in the text, which I've changed to 'b'.
Screenshot 2020-07-18 at 9 31 20 AM
Screenshot 2020-07-18 at 9 31 15 AM


frankharkins:
@Sahel13 good spot, thank you!

PR #541, opened 2020-07-18
Minor fixes in section 2.1 & 2.4

### LINKED ISSUE #432: The possiblity of some mistakes in Sec 2.5
## Describe the Issue
(1) " 𝑒^𝑖𝛼𝐴𝑍𝐵𝑍𝐶=𝑈" in text, but the circuit picture is using V.
(2) In the circuit picture of Toffoli, the last part of circuit is swapped in q1 and q2, but the text doesn't explain why it is swapped. 
(3) Text says "So far, we only have the power to do these arbitrary rotations around one axis. For a second axis, we simply do the 𝑅𝑧(𝜋/4) and 𝑅𝑥(𝜋/4) rotations in the opposite order." but the circuit order is same; HTHT.

## Expected Solution
(1) V in the the picture should be changed to U.
(2) I guess the connectivity of qubits is the cause of this. I think more explanation should be included. 
(3) The circuit might be changed to THTH.

realistix30:
(1) " 𝑒^𝑖𝛼𝐴𝑍𝐵𝑍𝐶=𝑈" in text, but the circuit picture is using V.
(2) In the circuit picture of Toffoli, the last part of circuit is swapped in q1 and q2, but the text doesn't explain why it is swapped.
(3) Text says "So far, we only have the power to do these arbitrary rotations around one axis. For a second axis, we simply do the 𝑅𝑧(𝜋/4) and 𝑅𝑥(𝜋/4) rotations in the opposite order." but the circuit order is same; HTHT.

Fixes #432 

## What Changes Were Made?
Added note together with Tofolli to explain the qubits connections in the device.
Changed U to V so that the text would match the example picture.
Reordered the circuit in section 2.5 based on the text. From HTHT -> THTH

## How Was This Tested?
Checked output in Jupyter Notebooks


ivanightingale:
For the 3rd change, I'd recommend running the whole notebook again so that the plot of the circuit is also corrected (to THTH).

PR #557, opened 2020-07-19
Grammar Fixes

spencerchurchill:
## What Changes Were Made?

I fixed a few clauses with incorrect comma placement and diction.


content/ch-states/atoms-computation.ipynb
```diff
@@ -6609,9 +6609,9 @@
...
     "\n",
-    "The half adder contains everything you need for addition. With the NOT, CNOT and Toffoli gates, we can create programs that add any set of numbers of any size.\n",
+    "The half adder contains everything you need for addition. With the NOT, CNOT, and Toffoli gates, we can create programs that add any set of numbers of any size.\n",
     "\n",
-    "These three gates are enough to do everything else in computing too. In fact, we can even do without the CNOT. Additionally, the NOT gate is only really needed to create bits with value ```1```. The Toffoli gate is essentially the atom of mathematics. It is the simplest element into which every other problem-solving technique can be compiled.\n",
+    "These three gates are enough to do everything else in computing too. In fact, we can even do without the CNOT. Additionally, the NOT gate is only really needed to create bits with value ```1```. The Toffoli gate is essentially the atom of mathematics. It is the simplest element in which every other problem-solving technique can be compiled.\n",
```
frankharkins, line 6614: Not sure about this one, maybe:

> It is the simplest element, from which every other problem-solving technique can be compiled.

?

PR #560, opened 2020-07-20
Fixissue425branch

### LINKED ISSUE #425: Wrong subtitle highlight
## Describe the Issue

When going to the last block of the page, if it is small it is not highlighted on the table of contents on the right. For example:

Link to "Contributors": https://qiskit.org/textbook/preface.html#Contributors
Result: "Contact" highlighted
Screen Shot 2020-06-03 at 10 04 23

## Expected Solution

"Contributors" highlighted

It only happens when the last block is not enough big and the previous block is also shown. See these examples:

- Small block ❌
Screen Shot 2020-06-03 at 10 11 00

- Big block ✅
Screen Shot 2020-06-03 at 10 14 14

anto-pv:

Fixes #
Tried to fix #425..but found bug
Modified the code definitions
## What Changes Were Made?
In the widgets/__init__.py the function declaration fro 
## How Was This Tested?

### DIFF
--- a/qiskit-textbook-src/qiskit_textbook/widgets/__init__.py
+++ b/qiskit-textbook-src/qiskit_textbook/widgets/__init__.py
@@ -39,7 +39,7 @@ def binary_widget(nbits=5):
     display(vbox)
     display(output.widget)
 
-
+#For the vector representation
 def state_vector_exercise(target):
     output = _pre()
     button = widgets.Button(description="Check", layout=widgets.Layout(width='5em'))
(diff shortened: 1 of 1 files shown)


frankharkins:
@anto-pv Firstly, this does not fix #425, but you have linked the issues through github's keywords and I do not think I can merge this without incorrectly closing the issue. I will have to close this PR.

I do not see a bug, from the changes it looks like you're adding some comments to the code. Commenting code is great where the code is not self-explanatory. Comments like these should go inside the function (first line after `def ...()`), and tell use something non-obvious about the code, maybe the chapter / section in which it is used could be useful.

Finally, you have unindented a definition used only inside another function. This should remain indented so it is not accessible to other functions. I think convention is to add an underscore to the start of the function to indicate this.

PR #566, opened 2020-07-20
Remove left-hand-sides for successive lines of equation

ivanightingale:
## What Changes Were Made?
Remove the left-hand-sides in successive lines of equation, at multiple places. This is to maintain a style consistent to the other notebooks and the more understandable practice in mathematical expressions.
##### Before the fix
image
##### After the fix
image


ivanightingale:
@frankharkins Could you please take a moment to review this?

PR #568, opened 2020-07-21
Fix Error in Code Blocks Caused by "Draw()" in Chapter 1.3 Representing Qubit States

PamelaPajarillo:
Issue with running code shown below
image
Fixes #your_issue_number_here

## What Changes Were Made?
Replaced qc.draw() with qc.draw('text') on line 468, 1091, 2877

## How Was This Tested?
image


frankharkins:
I have been avoiding this for a while since it is a qiskit bug and not a textbook bug. I think this solution works, but we should add a short comment explaining `text` is required due to a bug in qiskit. This commit should also be reverted after the bug is fixed in the next qiskit release.

PR #583, opened 2020-07-23
Matplotlib 3.3.0 Update: arg s => text

spencerchurchill:

While using Matplotlib 3.3.0:

```
MatplotlibDeprecationWarning: The 's' parameter of annotate() has been renamed 'text' since Matplotlib 3.3; support for the old name will be dropped two minor releases later.
  app.launch_new_instance()
```

This pull request will update the notebook to the most recent Matplotlib standards.

## What Changes Were Made?

I renamed the "s" argument to "text"

## How Was This Tested?

Before update:
Screen Shot 2020-07-23 at 9 22 01 AM

After update:
Screen Shot 2020-07-23 at 9 22 24 AM

Screen Shot 2020-07-23 at 9 25 49 AM

Pull request will break period finding as the "text" argument throws an exception
``` TypeError: annotate() missing 1 required positional argument: 's'```

I think it's best to let this request wait until matplotlib 3.3.0 is more standard.

### DIFF
--- a/content/ch-algorithms/shor.ipynb
+++ b/content/ch-algorithms/shor.ipynb
@@ -1233,8 +1233,8 @@
     "       title=\"Example of Periodic Function in Shor's Algorithm\")\n",
     "try: # plot r on the graph\n",
     "    r = yvals[1:].index(1) +1 \n",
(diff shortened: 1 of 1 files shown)


frankharkins:
Awesome! Thank you

PR #588, opened 2020-07-23
Likely typo: "all matrices" should be "all unitary matrices"

ivanightingale:
## Changes made
In Section 5 of 2.5-"Proving Universality", the sentence "since we have shown previously that **all matrices** can be expressed in this way, this is sufficient to show that we can reproduce all unitaries" should mean that **all unitary matrices** can be expressed in the form e^(iH).

Additionally, a small typo "an set of basic gates" is changed to "a set of basic gates".


content/ch-gates/proving-universality.ipynb
```diff
@@ -386,7 +386,7 @@
...
     "\n",
-    "Methods of this form can reproduce any unitary $U = e^{iH}$ for which $H$ can be expressed as a sum of tensor products of Paulis. Since we have shown previously that all matrices can be expressed in this way, this is sufficient to show that we can reproduce all unitaries. Though other methods may be better in practice, the main concept to take away from this chapter is that there is certainly a way to reproduce all multi-qubit unitaries using only the basic operations found in Qiskit. Quantum universality can be achieved!\n",
+    "Methods of this form can reproduce any unitary $U = e^{iH}$ for which $H$ can be expressed as a sum of tensor products of Paulis. Since we have shown previously that all unitary matrices can be expressed in this way, this is sufficient to show that we can reproduce all unitaries. Though other methods may be better in practice, the main concept to take away from this chapter is that there is certainly a way to reproduce all multi-qubit unitaries using only the basic operations found in Qiskit. Quantum universality can be achieved!\n",
```
HwajungKang, line 389: I think the original sentence is right.

PR #596, opened 2020-07-24
Equation typo

MahatKC:
Equation typo made the result show "\x | psi" instead of just the letter "x" and the symbol "psi".

This typo comes from PR #575, which switched "x" and "psi" but left the "\\" in its old place.


MahatKC:
@frankharkins I've checked the textbook page (https://qiskit.org/textbook/ch-states/representing-qubit-states.html#rules-measurement) and even though the merge has already happened, the error is still there... Is it a merge issue?

Issue #614, opened 2020-07-28
Add outer product and tensor product to linear algebra section


ryanfmandelbaum:
**Is your feature request related to a problem? Please describe.**
Currently, the textbook introduces the outer product and tensor products without explaining them. 

**Describe the solution you'd like**
I would like to add a section on the outer and tensor products to the linear algebra section.

**Describe alternatives you've considered**

**Additional context**
I have already discussed these changes with Frank and Abe.

PR #615, opened 2020-07-28
8.1 : Add section on tensor and outer products

ryanfmandelbaum:
# Changes made

Wrote new section on tensor and outer products with examples

# Justification

Discussed this addition with Abe and Frank already, filed it in issue #614


aasfaw:
fixes #614

frankharkins:
I see potential confusion here. We have:

* outer product |a> and |b> = |a><b| = [square matrix]
* outer product is specific case of tensor product
* tensor product |a> and |b> = |a>|b> = [column vector]

The key is we don't conj. transpose |b> in the second case, but this could be easily missed by a reader. Should maybe mention this explicitly?

PR #628, opened 2020-08-01
Replaced_two_hand_drawn_pictures

Aniruddha120:
Two hand drawn picture are replaced with a photoshoped image. Other images are difficult to understand, so only these two were attempted.

# Changes made

# Justification


frankharkins:
@Aniruddha120 thank you for your contribution. The images are already being redrawn professionally in the style of the textbook ([example of style](https://qiskit.org/textbook/ch-algorithms/images/superdense.jpg)). I have opened an issue (#644) to avoid this problem in the future.

PR #629, opened 2020-08-01
2.3, 1 Corrected the swapping of CNOT gate

kshhhv:
# Changes made
The 01 state is changed to 10.

# Justification

The CNOT gate swaps 10 and 11 qubits rather than 01 and 11.


kshhhv:
The little-endian convention of Qiskit was not taken into account.

PR #650, opened 2020-08-06
4.1.2 Changed 'state' to 'eigenstate' for better clarity

paniash:
# Changes made
Changed 'state' to 'eigenstate' for better clarity.

# Justification
$|\psi_min\rangle$ is the eigenstate of Hamiltonian H and hence the change makes better sense to the novice reader.


paniash:
Pardon me for the sloppy justification but apparently Github markdown doesn't support LaTeX.

PR #662, opened 2020-08-11
chapter 3.4: fix #660

### LINKED ISSUE #660: Deutsch-Jozsa Algorithm: Subscript
Hi, I have been studying DJ Algorithm from IBM Qiskit book and come across this problem. How do the subscripts turn into 10 or 11 from 1?

image

aditya-giri:
# Changes made
Modified the notation a little bit to avoid the confusion arising from indexing qubits by the register they are part of

# Justification
As pointed out in #660, the subscripts can get confusing (especially when we move from 1 to 2 subscripts without any indication as to what they represent)


content/ch-algorithms/deutsch-jozsa.ipynb
```diff
@@ -199,44 +199,44 @@
...
     "   <li> Apply Hadamard on all qubits\n",
     "    \n",
     "\n",
-    "$$\\lvert \\psi_1 \\rangle = \\frac{1}{2} \\left( \\lvert 0 0 \\rangle_1 + \\lvert 0 1 \\rangle_1 + \\lvert 1 0 \\rangle_1 + \\lvert 1 1 \\rangle_1 \\right) \\frac{1}{\\sqrt{2}} \\left( \\lvert 0 \\rangle_2 - \\lvert 1 \\rangle_2 \\right)  $$\n",
+    "$$\\lvert \\psi_1 \\rangle = \\frac{1}{2} \\left( \\lvert 0 0 \\rangle + \\lvert 0 1 \\rangle + \\lvert 1 0 \\rangle + \\lvert 1 1 \\rangle \\right) \\otimes \\frac{1}{\\sqrt{2}} \\left( \\lvert 0 \\rangle - \\lvert 1 \\rangle \\right)  $$\n",
     "\n",
     " \n",
     "   </li>\n",
     "    \n",
     "   <li> The oracle function can be implemented as $\\text{Q}_f = CX_{1a}CX_{2a}$, \n",
```
frankharkins, line 215: Can we explain the subscripts here please? I think `1a` means control is qubit 1, target is auxillary qubit (called output qubit in the rest of the article)? But this wasn't immediately obvious to me.

aditya-giri: Line 215? Yeah, I missed that sorry. I'll try and make it clearer. It's pretty vague right now what the oracle does and how it maps states.

aditya-giri: But first:
The example says `Let's go through a specific example for a two bit balanced function, where we apply X-gates before both CNOTs`. However, the X gates aren't explicitly applied anywhere. I know that applying X gates on both qubits when they're already in the |++> state is equivalent to doing nothing but maybe we should reword that?

frankharkins: hmm yes that is strange. Without working through the equation properly, I can't see it has any effect on the maths either, maybe we should just remove the line?

aditya-giri: Yeah I think we can safely remove it. Let me just take a look at it in its entirety once and ensure that everything makes sense.

aditya-giri: @frankharkins I've changed it a little, let me know your thoughts!

content/ch-algorithms/deutsch-jozsa.ipynb
```diff
@@ -193,50 +193,52 @@
    "source": [
     "## 2. Worked Example  <a id='example'></a>\n",
     "\n",
-    "Let's go through a specific example for a two bit balanced function, where we apply X-gates before both CNOTs.\n",
+    "Let's go through a specific example for a two bit balanced function: \n",
     "\n",
     "<ol>\n",
     "    <li> The first register of two qubits is initialized to $|00\\rangle$ and the second register qubit to $|1\\rangle$ \n",
+    "        \n",
+    "(Note that we are using subscripts 1, 2, and 3 to index the qubits. A subscript of \"12\" indicates the state of the register containing qubits 1 and 2)\n",
```
frankharkins, line 201: Should this maybe be 0, 1, 2 instead of 1, 2, 3?

content/ch-algorithms/deutsch-jozsa.ipynb
frankharkins: Looks good thank you!

content/ch-algorithms/deutsch-jozsa.ipynb
aditya-giri: Hmm I don't see a problem really, as long as it's consistent throughout the example. In fact 0-indexing is sometimes unintuitive for readers from other backgrounds. But that's your call to take :) I can change it if you feel that would be better.

frankharkins: Let's leave it 🙂

aditya-giri: Great 🙂

Issue #635, opened 2020-08-03
Add short explanation of quantum annealing


frankharkins:
**Is your feature request related to a problem? Please describe.**
Quantum annealing is a talked-about area of quantum computing and it should be acknowledged in the textbook. A chapter is out of scope (see #532) but readers are likely to come across the term and will want to know how it relates to what they're learning.

**Describe the solution you'd like**
A short paragraph summarising the concept of adiabatic QC / quantum annealing, how it relates to the gate based quantum computation model and maybe even a few proposed advantages / disadvantages. I believe this paragraph would fit well at the end of [2.5 Proving Universality](https://qiskit.org/textbook/ch-gates/proving-universality.html#5.-Proving-Universality-) 

**Additional context**
See #532 for related discussion.

charmerDark:
I think I could take up this and give you a draft in about two days.

frankharkins:
@charmerDark thank you!

charmerDark:
> In fact, the gate model of quantum computing is not the only universal model of quantum computation. There exist other forms of quantum computing such as the adiabatic quantum computing or the measurement based quantum computing. This means that it has been proven that there is a polynomial time and resource mapping from the  circuit gate model to these models of computing.These often leverage other quantum mechanical properties also in order to perform their computation.

> Quantum Annealing is another quantum computing model which is not universal nevertheless has some interesting applications.This means it cannot run the Shor's ,Grover's and many other quantum algorithms. It involves encoding problems into an energy landscape of sorts and then letting a wave function evolve into the landscape. While normal waves may get trapped at a local optima, quantum tunneling effect ensures the wave function reaches the global optima or really close levels. This can help with many optimization problems like the traveling salesman problem. 
> 
I was thinking of this for the paragraph. Please do let me know if this comes out as too popular science-ey or lacking in detail.

frankharkins:
@charmerDark thank you!

> In fact, the gate model of quantum computing is not the only universal model of quantum computation.

Need to mention the gate model is the model we've been using (not sure if that's mentioned earlier)

> There exist other forms of quantum computing such as the adiabatic quantum computing or the measurement based quantum computing.

Need to make clear to the reader that they do not need to understand adiabatic quantum / MBQC computing, just that these other forms exist.

For the second paragraph, I think maybe start with something like "Other forms of quanutm computation are not universal, but are applicable to specfic applications. For example quantum annealing..."

[...]

PR #669, opened 2020-08-13
Added paragraphs on Quantum Annealing and other models of quantum computing

charmerDark:
# Changes made

Added two paragraphs to content/ch-gates/proving-universality.ipynb as discussed in  #635 discussing models of quantum computing and Quantum Annealing
# Justification

Necessity of a section on quantum annealing was discussed in  #635 and first draft brought up after some more consultation


content/ch-gates/proving-universality.ipynb
```diff
@@ -388,7 +388,12 @@
...
+    "This gate set is not the only one that can acheive universality. For example it can be shown that just the Hadamard and Toffoli are sufficient for universality. Multiple other gates sets have also been considered and been proven universal, each motivated by different routes toward acheiving the gates fault-tolerantly.\n",
+    "\n",
+    "Everything we have discussed in this chapter and most of this book follows the circuit model of computation. However, the circuit model is not the only universal model of quantum computation. There exists other forms of quantum computing such as the adiabatic quantum computing or the measurement based quantum computer. This means that it has been proven that there is a polynomial time and resource mapping from the circuit model to these models of computing.These often leverage other quantum mechanical properties also in order to perform their computation. However we do not need to delve these other forms for our purposes.\n",
```
frankharkins, line 393: > ...most of this book...

I would just say "this book" as I believe all of the textbook uses gate based QC

> This means that it has been proven...

I would say "The fact that they are universal means it has been proven..."

> However we do not need to delve these...

Should make it clear to the reader the reason we have mentioned this. I imagine something like "It is important to be aware that these other forms of quantum computation exist, and that the benefits of each are hardware related. Since a universal model of quantum computation can perform any quantum algorithm, we need only stick with the gate model and can ignore other universal models."

frankharkins:
@charmerDark thanks for this. I have left a couple more comments, let me know what you think.

charmerDark:
Thanks a lot @frankharkins 
Wouldn't know what to do without you!
I've put in some more changes as we discussed. Please do let me know what you think!

frankharkins:
@charmerDark Thank you!

charmerDark:
> 
> 
> @charmerDark Thank you!

No @frankharkins Thank You! You carried me through the whole process! Sorry for all the trouble!

frankharkins:
@charmerDark no, I only guided. No trouble at all.

PR #674, opened 2020-08-14
typos fixed

shovitraj:
Typos and punctuations fixed

# Changes made

# Justification


shovitraj:
Fixed typo in Teleportation file and some punctuations in Shor's algorithm file.

frankharkins (changes requested):
With the Shor chapter, this looks like a 3000 line diff to add a comma. I know this is because you have used png instead of svg to re-render the outputs, but since this change does not affect the cell output I would prefer if you reset the file and added one comma, especially since svg is the preferred format for images in the textbook.

Issue #659, opened 2020-08-10
Model Enhancement for Hybrid Quantum-classical Neural Networks in the Qiskit Textbook


cyx617:
## Is your feature request related to a problem? Please describe.

[The chapter 4.1.5 of the Qiskit textbook](https://qiskit.org/textbook/ch-machine-learning/machine-learning-qiskit-pytorch.html) shows a very intriguing example of hybrid quantum-classical neural networks with PyTorch and Qiskit and it demonstrates the ability of integrating Qiskit and PyTorch in the field of quantum machine learning. But some details of the code might need to be reconsidered. For example, when creating the hybrid neural network, we use the code below
```
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.conv1 = nn.Conv2d(1, 32, kernel_size=5)
        self.conv2 = nn.Conv2d(32, 64, kernel_size=5)
        self.dropout = nn.Dropout2d()
        self.fc1 = nn.Linear(256, 64)
        self.fc2 = nn.Linear(64, 1)
        self.hybrid = Hybrid(qiskit.Aer.get_backend('qasm_simulator'), 100, np.pi / 2)

    def forward(self, x):
        x = F.relu(self.conv1(x))
        x = F.relu(self.conv2(x))
        x = F.max_pool2d(x, 2)
        x = self.dropout(x)
        x = x.view(-1, 256)
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        x = self.hybrid(x)
        return torch.cat((x, 1 - x), -1)

```
[...]

PR #684, opened 2020-08-17
Chapter 4.1.5: fix incorrect dimensions of tensors in hybrid quantum-classical neural network 

cyx617:
# Changes made

Modified model architecture of the hybrid quantum-classical neural network, as pointed out in [#659](https://github.com/qiskit-community/qiskit-textbook/issues/659)

# Justification

In the original model, the dimensions of the input tensors for fc1 and fc2 layers are inconsistent with the batch size set for model training, which results in long training time and fluctuating loss curve. After the model is updated, it can achieve a better performance in terms of both train/test loss and running time.


frankharkins (approved):

frankharkins:
Thank you! For the record, @amyami187 has approved this (I couldn't seem to request the review through github).

PR #712, opened 2020-09-01
1.5 Correction in the definition of Big O notation

abhrac:
# Changes made
Corrected the relation specified between f(x) and g(x) in the definition of Big O notation.

# Justification
f(x) = O(g(x)) implies that f(x) is upper bounded by Mg(x), not the other way round.
https://en.wikipedia.org/wiki/Big_O_notation#Formal_definition


content/ch-states/case-for-quantum.ipynb
```diff
@@ -59,7 +59,7 @@
     "  <summary>Definition: Big O notation (Click to expand)</summary>\n",
     "For some example functions $f(x)$ and $g(x)$ and parameter $x$, the statement $f(x) = O(g(x))$  means that there exists some finite numbers $M>0$ and $x_0$ such that\n",
```
3yakuya, line 60: Could we update this to "exist" rather than exists, since we are referring to numbers?

3yakuya (dismissed):
One minor comment that would be good to improve here, important catch otherwise, could be very confusing for readers unfamiliar with complexity theory!

content/ch-states/case-for-quantum.ipynb
abhrac: Addressed.

frankharkins (approved):
Looks good, thank you!

Issue #721, opened 2020-09-02
Typo in the quantum teleportation circuit in chapter 3.1


wmurphy-collabstar:
**Describe the issue**
In the last part of the circuit, when the X gate and then the Z gate are applied to the third qubit, the classical registers are labelled 'z_meas' where the X gate is applied and 'x_meas' where the Z gate is applied. 
**Describe what you expect to see**
I expect to see the names of the classical registers flipped so that the X gate corresponds to the x_meas register and the Z gate corresponds to the z_meas register.

paniash:
@wmurphy-collabstar Thank you for pointing this out. I have made the changes and will make a PR for it.

wmurphy-collabstar:
Thank you very much @paniash

frankharkins:
This was fixed in #724

PR #724, opened 2020-09-03
3.1 Fixed labels in teleportation circuit diagram

paniash:
# Changes made

Fixes issue #721 

# Justification

The X measurement was being done on 'z_meas' classical register and Z measurement was being done on 'x_meas' classical register.


paniash:
@frankharkins Apologies for the commits related to #722. I'm still new to Git. I'll make sure to make changes in separate branches from here onwards.

frankharkins (approved):

PR #725, opened 2020-09-03
ch-appendix/linear_algebra: Unitary matrices do not preserve the state ch-states: avoided confusing language ch:single qubit gates

rupeshknn:
Unitary matrices preserve the length/magnitude/probability of the quantum state and not the state itself

# Changes made

replaced "Quantum state" with "magnitude or length of quantum state"


content/ch-appendix/linear_algebra.ipynb
```diff
@@ -4454,7 +4454,7 @@
     "\n",
     "<br>\n",
     "\n",
-    "The reason unitary matrices are important will become more apparent in the section on Hilbert spaces, and more so in the quantum mechanics subtopic of this textbook.  The basic idea is that evolution of a quantum state by application of a unitary matrix \"preserves\" the quantum state.\n",
+    "The reason unitary matrices are important will become more apparent in the section on Hilbert spaces, and more so in the quantum mechanics subtopic of this textbook.  The basic idea is that evolution of a quantum state by application of a unitary matrix \"preserves\" the magnitude or length of the quantum state.\n",
```
3yakuya, line 4457: Is there a double space after the first sentence?

As for the word itself, perhaps we could go with "norm" rather than length, since the latter can be confusing (length can also be interpreted as number off possible measured states, number if qubits etc)

3yakuya (changes requested):
Definitely a great catch and important to fix, but I think we can come up with even better wording.

content/ch-appendix/linear_algebra.ipynb
rupeshknn: yes, I agree, the norm would the better word

frankharkins (approved):

rupeshknn:
hey Travis CI / Travis CI - Pull Request failed, what do I do? @frankharkins

frankharkins (approved):

PR #727, opened 2020-09-04
1.3 Fix + flipped to - in measurement calculation

Iambecomeroot:
I believe there is a typo in the calculations explaining how to measure a qubit. The + sign becomes a - sign on the next line seemingly for no reason.

# Changes made
Change - for +

# Justification
The + sign becomes a - sign on the next line seemingly for no reason.


paniash:
Looks like this change is implemented in #722.

Iambecomeroot:
Sorry

PR #729, opened 2020-09-04
Added an analogy to chapter 3.2

wmurphy-collabstar:
# Changes made

I added an analogy about the quantum teleportation protocol being similar to a friend sending roses to another friend, and added a comparison to the protocol at each step of the analogy.

# Justification

I initially had a hard time understanding the protocol the way it is, and I thought other people might feel the same way. That is why I added this analogy as a way to help others understand the quantum teleportation protocol. I am very open to all criticism, such as whether this analogy really fits this chapter, there is a better place to put it, the wording could be changed, images should be added, there should be more edits in general, etc.

### DIFF
--- a/content/ch-algorithms/teleportation.ipynb
+++ b/content/ch-algorithms/teleportation.ipynb
@@ -20,16 +20,17 @@
     "## Contents\n",
     "\n",
     "1. [Overview](#overview)    \n",
-    "2. [The Quantum Teleportation Protocol](#how)       \n",
-    "3. [Simulating the Teleportation Protocol](#simulating)   \n",
-    "    3.1 [How will we Test this Result on a Real Quantum Computer?](#testing)   \n",
-    "    3.2 [Using the Statevector Simulator](#simulating-sv)     \n",
-    "    3.3 [Using the QASM Simulator](#simulating-qs)  \n",
-    "4. [Understanding Quantum Teleportation](#understanding-qt)\n",
-    "5. [Teleportation on a Real Quantum Computer](#real_qc)    \n",
-    "    5.1 [IBM hardware and Deferred Measurement](#deferred-measurement)    \n",
-    "    5.2 [Executing](#executing)    \n",
-    "6. [References](#references)\n",
+    "2. [The Quantum Teleportation Protocol](#how) \n",
+    "3. [Jacklyn sends Vivian Roses: An Analogy](#analogy)\n",
+    "4. [Simulating the Teleportation Protocol](#simulating)   \n",
+    "    4.1 [How will we Test this Result on a Real Quantum Computer?](#testing)   \n",
+    "    4.2 [Using the Statevector Simulator](#simulating-sv)     \n",
+    "    4.3 [Using the QASM Simulator](#simulating-qs)  \n",
+    "5. [Understanding Quantum Teleportation](#understanding-qt)\n",
+    "6. [Teleportation on a Real Quantum Computer](#real_qc)    \n",
+    "    6.1 [IBM hardware and Deferred Measurement](#deferred-measurement)    \n",
+    "    6.2 [Executing](#executing)    \n",
+    "7. [References](#references)\n",
     "\n"
    ]
   },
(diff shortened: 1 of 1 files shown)


content/ch-algorithms/teleportation.ipynb
```diff
@@ -2454,14 +2455,41 @@
...
-    "## 3. Simulating the Teleportation Protocol <a id='simulating'></a>"
+    "## 3. Jacklyn sends Vivian Roses: An Analogy <a id='analogy'></a>"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### 3.1 How Will We Test the Protocol on a Quantum Computer? <a id='testing'></a>"
+    "Now that we know the quantum teleportation protocol, I want to take a step back and make the protocol more relatable with an analogy.\n",
```
3yakuya, line 2465: Let's try and stick with the multi-author language. Perhaps "let's take a step back and..."

content/ch-algorithms/teleportation.ipynb
```diff
@@ -2454,14 +2455,41 @@
...
+    "\n",
+    "Jacklyn, the first qubit, wants to send roses to Vivian. Those roses are the information stored in the first qubit. In order to send the roses, she gets help from FlowerDelivery, a delivery company, to deliver the roses to Vivian. The company provides the delivery truck and Vivian's information, and together those pieces of information help determine the directions to Vivian's house. The entangled qubit channel is this route, while the entangled qubits are the delivery truck and Vivian.\n",
```
3yakuya, line 2467: I think it's rather confusing that in the original protocol we have Alice _send qubits_ to Bob, while here _Jacklyn and Vivian are the qubits_ trying to send each other information.

3yakuya (changes requested):
I think a good analogy can be helpful indeed, but in this case the fact that Jacklyn and Vivian are qubits, rather than people trying to send qubits, gets very confusing.

Roses are being sent in the analogy, but they are not analogous to sent qubits/bits but rather information stored in the qubit - I found it not the easiest to grasp myself.

Could we rewrite the analogy to make it a bit more concise and consider elements to be analogous (who is Alice in this analogy? Who is Bob? What is the quantum communication channel? Etc.)

content/ch-algorithms/teleportation.ipynb
wmurphy-collabstar: I will incorporate that, thank you.

content/ch-algorithms/teleportation.ipynb
wmurphy-collabstar: I agree and am willing to make the analogy more relatable to the Quantum Teleportation Protocol. Thank you for the comment.

frankharkins:
@wmurphy-collabstar thanks for contributing and trying to help out new learners that follow in your footsteps. I am always wary of comparisons to classical scenarios, and I think this could potentially be misleading:

- It sounds like the the purpose of applying the gates at the end is to account for errors in the system, which is not true as we can measure |0>|1> on perfect hardware.

- I assume with this analogy that roses = quantum information, and the delivery van is the third party (Telamon). Roses in the delivery van sounds like there is a direct channel of quantum communication between Alice and Bob. I think Alice giving the roses to the delivery van is also misleading as, after recieving the entangled qubit, Alice needs no more contact with Telamon. Alice does not even have to have her qubit state at that point.

An explanation of the protocol has recently been added ([link](https://qiskit.org/textbook/ch-algorithms/teleportation.html#4-understanding-quantum-teleportation)), do you think this could have gone some of the way to resolving your confusion when you first came across the protocol?

wmurphy-collabstar:
I understand the hesitation with the analogy and I've realized that my analogy is misleading. I am also fine with the idea of not adding the analogy if adding it would create more confusion than clarity, but there is still a question I have about the algorithm. What I can't wrap my head around is conceptually why is the quantum teleportation algorithm the way it is? What I mean is, why do we apply the gates that we do, and why would we not do it another way? I understand it mathematically, but I'm wondering if there is a more conceptual way of explaining the purpose of each gate being applied in the algorithm. If you have any resources that would help with my questions, or a place to go, that would be greatly appreciated.

wmurphy-collabstar:
Also, thank you for your feedback. I really appreciate it.

frankharkins:
@wmurphy-collabstar with a Bell pair (e.g. |00> + |11>), when we measure one qubit, we know what the state of the other will be (measure one as |0> -> other is |0>). If you and I have either side of a Bell pair, if I measure |0> but want you to have the state |1>, I can then tell you to apply an X-gate to your qubit and we know you'll then have the state |1>. Teleportation is the same idea, but we instead measure the qubits in such a way that the recipient's qubit is collapsed into some state containing 'a' and 'b'. We can then use the results of the measurement to recitfy the recipient's qubit to the state a|0> + b|1>.

Maybe this SO answer will help: https://physics.stackexchange.com/a/138501/120557

wmurphy-collabstar:
Thank you so much for the explanation and the link for more information. I think I now understand the quantum teleportation protocol much more than before. I'm going to postpone this pull request now, and if I think of an analogy that can take into account all of the comments given here, I'll make a new pull request. Thanks again for the help!

frankharkins:
@wmurphy-collabstar it's a pleasure!

PR #730, opened 2020-09-05
2.3 Rephrasing a portion in the explanation of Phase Kickback

abhrac:
# Changes made
Rephrasing a portion in the explanation of Phase Kickback.

# Justification
The explanation of the CNOT identity which induces a phase kickback is ambiguous. More specifically, the following portion:
"this phase factor applies to the target qubit only when it is in the state  |1⟩"
It is not clear which qubit is referred to by "it" in the above statement. Also, neither of the qubits are completely in state |1⟩ (since both are in superposition), as depicted in the example under consideration.
This PR is to make what the "it" in the statement refers to, more explicit.


paniash (changes requested):
The important point about this particular example is when the control qubit is in the state |1>. So changing it to |-> does not make sense. However, the way the sentence is put can be confusing to beginners and could be changed as:

"The interesting effect is when our control qubit is in superposition, this phase factor applies to the target qubit only when the control qubit is in the state |1⟩, which adds a relative phase to our control qubit"

abhrac:
@paniash I have some concerns regarding the correctness of the statement you have suggested. But before I begin, let me just clarify a fact in order to ensure that we are on the same page regarding which qubit is control and which one is target, and how each one gets transformed by the CNOT.

According to Qiskit's convention, the more significant qubit appears first (little-endian). So, the order is |qubit-1, qubit-0>. In the example under consideration, qubit-1 is the target qubit and qubit-0 is the control qubit.
The initial state of the system was |-+>, which gets transformed to |-->, due to phase kickback. Phase kickback implies that the target qubit remains unchanged and the control qubit suffers a phase shift. Here, the target qubit remains in its initial state, |->, all along. The control qubit shifts from |+> to |->.

Now I proceed with enumerating my concerns.
My first concern is regarding the portion: "this phase factor applies to the target qubit only when the control qubit is in the state |1⟩". Here they are:
- Prior to being put into superposition, the control qubit was in the state |0>, because of which it was transformed into |+> after a Hadamard was applied. So, the control qubit was never |1> during the entire process.
- However, the target qubit, prior to being put into superposition, was in the state |1>, because of which it was transformed into |-> after a Hadamard was applied.

My next concern is regarding the following portion of your comment: "The important point about this particular example is when the control qubit is in the state |1>.", which are as follows:
- The control qubit was never in the state |1> during the entire process, as I have pointed out above.
- Phase kickback for the CNOT gate does not apply to qubits in the computational basis, since in such a basis, the phase added is global in nature and has no physical significance. Hence, I don't think the point that the author tries to make in this statement applies to either of the computational basis states (|0> or |1>).

paniash:
@abhrac Thank you for the detailed comment. I am aware of the way Qiskit orders its qubits and still stand by my point.

> Phase kickback for the CNOT gate does not apply to qubits in the computational basis, since in such a basis, the phase added is global in nature and has no physical significance. 
Yes, this is correct. It only applies when the control qubit (the rightmost qubit in Qiskit's convention) is in a superposition and not when in states |0> or |1> alone.

Having said that, this superposition in the control qubit implements kickback when a CNOT is applied. The phase change in the target qubit occurs only when the control qubit is in the state |1> which is true when the control is in the state |+> since |+> = (|0> + |1>)/sqrt(2).

pic-selected-200906-0157-22

Notice that the control qubit is |+>. The phase change of the target occurs in a part of the control qubit (|1>). Hence, giving it a relative phase shift of the entire control qubit and thus essentially rendering a change of the state of the control qubit, which is the kickback. 

Hopefully, this makes it clear. :)

abhrac:
@paniash Many thanks for the clarification. I now clearly understand the intention with which you are making your point. I completely agree with the description you have provided in your comment.

However, I would still say that the statement you had suggested was somewhat ambiguous and does not entirely reflect your intentions (which is why I had my initial concerns). So, I suggest a few modifications on top of your suggested statement, which are as follows:

"The interesting effect is when our control qubit is in superposition, this phase factor **is applied to** the target qubit** only by the component of the control qubit that lies in the direction of** |1⟩, which **in turn kicks back** a relative phase to our control qubit."

Kindly let me know if this works.

paniash:
@frankharkins I think you should have a look at this.

paniash:
@abhrac I think a better person to approve of this would be @frankharkins.

abhrac:
@paniash Sure, but can you let me know if I have been able to capture the point that you are trying to convey in the sentence above? If so, then I will go ahead and make the change and update the PR title and description.

paniash:
> "The interesting effect is when our control qubit is in superposition, the component of the control qubit that lies in the direction of |1⟩ applies this phase factor to the target qubit, which in turn kicks back a relative phase to our control qubit."

Yes, this does convey my point. Very well put!

abhrac:
@paniash Thanks for confirming. And also, many thanks for suggesting the correction :)

frankharkins (approved):

PR #735, opened 2020-09-08
2.3 Fix typo (rightmost -> leftmost)

Iambecomeroot:
# Changes made
Replace "rightmost" with "leftmost"

# Justification
The text says that the rightmost qubit is rotated, but in the picture
above, it is clear that the leftmost qubit is the one that is rotated.


Iambecomeroot:
Maybe we could clarify further by putting (qubit 0) in parenthesis after "leftmost"

frankharkins (approved):

PR #752, opened 2020-09-16
3.9 Fix minor issues in the text

theerfan:
# Changes made
I fixed one grammar issue, one capitalization issue and one issue in the less signs (which given the following summation formula, should've been \leq signs.)


frankharkins (approved):

frankharkins:
thank you!

Issue #769, opened 2020-09-25
Shots should be not 1024 but 10000 to much the context [Chapter 5.1]


YumaNK:
**Describe the issue**
The phrase  
"We'll repeat the process shots=1024 times to see how likely different results are."  
does not much the following phrase 
" In total, much less than 100 samples come out with a majority of 1s. When using this circuit to encode a 0, this means that P < 1%"

**Describe what you expect to see**

shots = 10000 should be correct. 

Then the cell before the phrase;
```
from qiskit import QuantumCircuit, execute, Aer

qc0 = QuantumCircuit(3,3,name='0') # initialize circuit with three qubits in the 0 state

qc0.measure(qc0.qregs[0],qc0.cregs[0]) # measure the qubits

# run the circuit with th noise model and extract the counts
counts = execute( qc0, Aer.get_backend('qasm_simulator'),noise_model=noise_model).result().get_counts()

print(counts)
```

should be

```
from qiskit import QuantumCircuit, execute, Aer

qc0 = QuantumCircuit(3,3,name='0') # initialize circuit with three qubits in the 0 state

qc0.measure(qc0.qregs[0],qc0.cregs[0]) # measure the qubits

# run the circuit with th noise model and extract the counts
counts = execute( qc0, Aer.get_backend('qasm_simulator'),noise_model=noise_model, shots=10000).result().get_counts()

print(counts)
```

tgag17:
Hi! Should the number of shots be changed or the percentage mentioned in the notebook?

YumaNK:
I guess shots should be changed.

tgag17:
Yes I checked by changing the shots and the cases with majority of '1's were generated which were not there with 1024 shots.
Can I fix this in the main notebook?

YumaNK:
Sure. Please fix it. Thanks

PR #771, opened 2020-09-26
5.1 Fixed issue769 by changing the number of shots

tgag17:
# Changes made

Fixes issue #769 
Changed the number of shots from default(1024) to 10000
# Justification

The text mentioned that the number of shots for a particular case are less than 100 making it less than 1% which is true for 10000 shots. Also the cases mentioned in the text were only generated with shots greater than 1024


frankharkins:
@tgag17 @YumaNK From what I understand of the issue (#769): The problem is that 100 is not approximately 1% of 1024, so the phrase:
> much less than less than 100 samples come out with a majority of 1s. When using this circuit to encode a 0, this means that P < 1%

is not correct? 

You seem to have changed the number of shots to match this statement, but you have not changed the line:

> We'll repeat the process shots=1024 times to see how likely different results are.

So either this needs changing too, or we keep the original code and simply change the line to read:

> much less than less than **10** samples come out with a majority of 1s. When using this circuit to encode a 0, this means that P < 1%

Please let me know if I have misunderstood this.

tgag17:
@frankharkins Sorry, I think I missed the line in the context. If we keep the original code then the cases with majority of 1s are not generated at all. So what do you recommend should be done

frankharkins:
That's ok right? As 0 << 10

tgag17:
Oh ok. I was also confused as to what changes should be made and thus discussed it before editing. It would be better to just change the 100 to 10 then.

tgag17:
Should this PR be closed and another one be opened?

frankharkins:
You can make the changes in this PR, or open another if it's easier for you.

PR #773, opened 2020-09-26
5.2 Fixed #731 and corrected code semantics

### LINKED ISSUE #731: "not invertible" might be misleading in measurement-error-mitigation
**Describe the issue**
In the chapter of measurement-error-mitigation, there is phrase "This time we find a more interesting matrix, and one that is not invertible."

If one read it in straightforward way,  "not invertible" will be understood that the matrix does not have inverse matrix.
However, this matrix do have inverse matrix as its determinant is not zero.

Considering the context, one may infer that this phrase means that the matrix cannot reproduce ideal result. 
It would be great if the phrase is rephrased in the way that one can clearly understand what it actually mean.

image

**Describe what you expect to see*

paniash:
# Changes made

**Before**: This time we find a more interesting matrix, and one that is not invertible.
**After**: This time we find a more interesting matrix, and one that we cannot use in the approach that we described earlier.

Also corrected some code semantics where qiskit is imported as `from qiskit import *`. But in some places, code such as 
```python
backend = qiskit.Aer.get_backend('qasm_simulator')
job = qiskit.execute(meas_calibs, backend=backend, shots=1000)
```

has been used instead of
```python
backend = Aer.get_backend('qasm_simulator')
job = execute(meas_calibs, backend=backend, shots=1000)
```
which is more popularly used.

I also changed `from qiskit import *` to
```python
from qiskit import QuantumRegister, QuantumCircuit, Aer, execute
```
for the sake of only importing modules that are actually used.

# Justification

As described in issue #731, I paraphrased the part where it says that the matrix is not invertible in the sense that the basic method of error mitigation cannot be used. The novice reader is likely to be confused, since one could take it to be that the inverse of the matrix does not exist, when really it is not the case.

### DIFF
--- a/content/ch-quantum-hardware/measurement-error-mitigation.ipynb
+++ b/content/ch-quantum-hardware/measurement-error-mitigation.ipynb
@@ -17,7 +17,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from qiskit import *"
+    "from qiskit import QuantumCircuit, QuantumRegister, Aer, execute"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


frankharkins (approved):

frankharkins:
Much better! Thank you

PR #789, opened 2020-10-19
6.2 Fixes deprecated use of pulse_lib and rel_minimum

HTransistor:
# Changes made
Replaced pulse_lib (deprecated) with the renamed equivalent (pulse.library).
There seems to be a change in the device as most of the results given back by amorak  give back plots with a reflection along the x-axis. As a result rel_minima does not find the right frequency, as it needs to find the maxima. The function and the according text passage was changed accordingly.

So having the same frequency (4.62GHz) the notebook now gives out a different optimal amplitude (0.37 instead of 0.44), resulting in a much worse separation between state` |1> `and  `|2>` (last plot). Maybe something on the device or the pulse library has changed?
6_2_different_amplitude
6_2_bad_higher_level

Maybe the hardcoded 8 is causing some issues?
hardcoded_8

# Justification
pulse_lib_is_library
6_2_inverted_plot
6_2_extrema_swapped


HTransistor:
Today I rerun the notebook in the changed version (as in this PR) and got a more sophisticated result.
 Not sure if it was a problem with cached variables... But everything seems to work fine now.
6_1_higher_states_work

frankharkins (approved):

PR #793, opened 2020-10-20
6.3 Fixes some signs and typos

HTransistor:
# Changes made
Some signs and typos

# Justification
This formula was used on the rest of equations:
6_3_interaction_picture
By the way I would be interested in the derivation of this equation, as it is not quite clear how the commutator relations and addition theorems are used here...

Because of the ladder operators being defined inverted, the equation in the end should be inverted, too, but it was not.
6_3_ladder_operators


nbronn (approved):
Thanks for fixing the typos! This looks good to me.

nbronn:
Also @HTransistor, I believe I calculated the above relations using the matrix representations of the Pauli operators, but it would be more instructive to do that from the commutation relations. I'll add this explanation soon.

frankharkins (approved):

Issue #795, opened 2020-10-20
Chapter 4.1.2


pbark:
**Describe the issue**
Last part of chapter 4.1.2 where the H_2 energy is estimated with noisy calculation the exact value is not taking into consideration the nuclear repulsion. This may be misleading for users with chemistry background that are not familiar with the qiskit drivers. 

**Describe what you expect to see*
The fix is relatively easy and only content change.

PR #796, opened 2020-10-20
4.1.2. fix in vqe molecules , add nuclear repulsion energy in H_2 molecule noisy calculations

pbark:
Affects chapter 4.1.2. where the exact energy of H_2 in noisy calculation was not also including the nuclear repulsion. 

When merged should close issue #795 

Should be fixed now

# Changes made

updated the printouts of the exact and vqe energies to include the nuclear repulsion 

# Justification

this may be misleading for people not familiar with qiskit drivers. Especially since this is inconsistent with LiH calculation of the same chapter.


frankharkins:
@pbark this diff is quite difficult to read. Can you tell me which lines you've changed?

content/ch-applications/vqe-molecules.ipynb
```diff
@@ -166,7 +166,11 @@
     "np.random.seed(999999)\n",
     "target_distr = np.random.rand(2)\n",
     "# We now convert the random vector into a valid probability vector\n",
-    "target_distr /= sum(target_distr)"
+    "target_distr /= sum(target_distr)\n",
+    "\n",
+    "import warnings\n",
+    "warnings.filterwarnings(\"ignore\", category=DeprecationWarning)\n",
```
frankharkins, line 172: Can we remove these lines that hide the deprecation warnings please? If the functions are deprecated we should update them (in another PR) instead of ignoring them as they will eventually break the notebook when the functionality is removed in a future release.

PR #797, opened 2020-10-21
6.4 Fixes typos and sign in delta

HTransistor:
# Changes made
Typo in a commutator was misplaced.
Sign in delta.

# Justification
Delta sign was changed as the equations imply it.

6_4_wrong_delta


HTransistor:
Actually the equation in first line is wrong [H_0, eta], because eta is wrong as H_0 was wrong. So the fix is just to change the - sign to + in H_0, when defining it in the python code.

content/ch-quantum-hardware/cQED-JC-SW.ipynb
```diff
@@ -154,7 +154,7 @@
     "  + \\frac{1}{2} \\left[ \\left[ H_0 + \\lambda(H_1+H_2), \\lambda S^{(1)}\\right], \\lambda S^{(1)}\\right]\n",
     "  + \\left[H_0 + \\lambda(H_1+H_2), \\lambda^2 S^{(2)}\\right] + \\ldots \\\\\n",
     "  \\approx H_0 + \\lambda \\left( H_1 + H_2 + \\left[H_0, S^{(1)}\\right] \\right) \n",
-    "  + \\lambda^2 \\left( \\left[H_1+H_2, S^{(1)}\\right] + \\frac{1}{2} \\left[ \\left[H_0, S^{(1)},\\right] S^{(1)}\\right] + \\left[H_0, S^{(2)}\\right]\\right)\n",
+    "  + \\lambda^2 \\left( \\left[H_1+H_2, S^{(1)}\\right] + \\frac{1}{2} \\left[ \\left[H_0, S^{(1)}\\right], S^{(1)}\\right] + \\left[H_0, S^{(2)}\\right]\\right)\n",
```
nbronn, line 157: Thanks for fixing the comma placement!

content/ch-quantum-hardware/cQED-JC-SW.ipynb
```diff
@@ -258,7 +258,7 @@
    ],
    "source": [
     "# define J-C Hamiltonian in terms of diagonal and non-block diagonal terms\n",
-    "H0 = wr*Dagger(a)*a - (1/2)*wq*sz; \n",
+    "H0 = wr*Dagger(a)*a + (1/2)*wq*sz; \n",
```
nbronn, line 261: This is correct here, the error is with the original Hamiltonain.

content/ch-quantum-hardware/cQED-JC-SW.ipynb
```diff
@@ -469,7 +469,7 @@
    "source": [
     "This is typically written as\n",
     "$$\n",
-    "H_{\\rm eff} = \\left(\\omega_r + \\frac{g^2}{\\Delta}\\sigma_z\\right)a^\\dagger a  - \\frac{1}{2}\\left(\\omega_q -\\frac{g^2}{\\Delta}\\right) \\sigma_z\n",
+    "H_{\\rm eff} = \\frac{g^2}{2\\Delta} \\left(\\omega_r + \\frac{g^2}{\\Delta}\\sigma_z\\right)a^\\dagger a  - \\frac{1}{2}\\left(\\omega_q -\\frac{g^2}{\\Delta}\\right) \\sigma_z\n",
```
nbronn, line 472: I think the original version is correct here.

nbronn (changes requested):
Thanks again for the fixes @HTransistor! I left some single comments on the PR, but basically the original (JC/RWA) Hamiltonian has a sign error on the qubit term, which should be $-\frac{1}{2} \omega_q \sigma_z$ because the |0> state has a lower energy than the |1> state which you can see from the definition of the z-Pauli matrix.

content/ch-quantum-hardware/cQED-JC-SW.ipynb
HTransistor: If the - is right, then the issue with the delta of the original post persists.
To be more precise: In the both equations there is no minus at all, so it can not imply the A and B with the definition of the given Delta (with an the minus).

content/ch-quantum-hardware/cQED-JC-SW.ipynb
HTransistor: Sorry forgot to take a screenshot. I think in the second equation the 1 in the brakets of the first equation is missing (clearly in eq1 are 5 terms, but in eq2 just 4).
6_1_missing_term

HTransistor:
Despite it sounds reasonable to have the - in the very first Hamiltonian, that does not fix the whole issue. I hope I clarified it in the subcomments.

content/ch-quantum-hardware/cQED-JC-SW.ipynb
nbronn: Thanks for the careful look @HTransistor! I see now that the problem is arising due to the \sigma_+ and \sigma_- being defined for spins in `sympy`, rather than qubit levels, so there is a discrepancy between the +/-. Then you are indeed correct that the A and B should have the minus signs. I am not sure what the most elegant solution for replacing the qubit raising and lowering operators is, however.

HTransistor: Alright, good that we find that one. I renamed the ladder operators (inversed, and made a comment in the code) and reverted the old change in H0, in the newest PR. Now it should be fine.

nbronn (approved):
Looks good to me!

frankharkins (approved):

PR #801, opened 2020-10-22
4.1.2 VQE: Updated Deprecated Code

frankharkins:
# Changes made
Removed suppression of deprecation warnings and updated the deprecated code to the latest version of qiskit.
Merging this will merge #796


pbark (changes requested):
Would it be possible to run it with the latest aqua before merging so that you get the latest deprecation warnings? The rest lgtm.

frankharkins:
@pbark thank you for pointing that out, I did not realise I did not have the most up to date qiskit installed. I have run the code with aqua 0.8.0 and noticed two more deprecation warnings (that I assume you were referring to). It looks like the deprecation warnings are generated by Aqua's internal code so we'll have to wait for the next release for the Aqua team to update their code.

pbark (approved):

PR #807, opened 2020-10-30
4.2 New Chapter Suggestion - Variational Quantum Regression WIP

clearya1:
# Changes made

Over the summer, I wrote an algorithm which performs polynomial regression using the Qiskit software. The algorithm is a variational one, in that it computes a cost function using quantum hardware, coupled with classical optimisers. 

This algorithm takes two data sets - an independent data set (x) and dependent data set (y) and attempts to return the line of best fit for an n-th order polynomial. The inputs to the algorithm are the data sets (x & y) and the polynomial order (n).

**What should the reader gain from this new content?**

After the content, the student will understand how to compute the inner product of two states. They will gain a deeper understanding of the principle of variational quantum computing, which has been discussed in Section 4 of the Qiskit textbook. Putting these two learnings together, they will learn a method to perform polynomial regression.

**Where do you expect this new content to fit in with the current content?**

I was thinking that this work might be a nice addition to the textbook under the Implementations of Recent Quantum Algorithms section (4.2), so I created a Jupyter notebook similar to the Qiskit textbook style. Everything can be found at this repo: [VQR](https://github.com/clearya1/Variational-Quantum-Regression.git)

**What material should this new content be based on?**

This work was inspired by the section 4.2.1 of the Qiskit textbook. 
The method for computing inner products is based on this [paper](https://arxiv.org/pdf/1902.10394.pdf). 
The variational computing aspect of the work is based on this [paper](https://arxiv.org/pdf/1909.05820.pdf).

**Are you proposing to write this new content?** (y/n)

I have already written the Jupyter Notebook, which can also be found as part of the full [repo](https://github.com/clearya1/Variational-Quantum-Regression.git). I am of course willing to edit the Jupyter notebook as appropriate!

# Justification

Justification for this new content suggestion can be found above in the *what should the reader gain from this new content* section.


content/ch-applications/Variational Quantum Regression.ipynb
```diff
@@ -0,0 +1,893 @@
...
+    "\n",
+    "\\begin{equation}\n",
+    "    P(0) = \\frac{1}{2}(1+Re[\\braket{x}{y}])\n",
+    "\\end{equation}\n",
+    "\n",
+    "This follows because:\n",
+    "\n",
+    "\\begin{align}\n",
+    "    P(0) &= |\\bra{0}\\otimes\\mathbb{1}\\ket{\\phi}|^2 \\\\\n",
+    "    &= \\frac{1}{4}\\braket{x+y}{x+y} \\\\\n",
```
frankharkins, line 82: I don't follow this step, can you explain please?

content/ch-applications/Variational Quantum Regression.ipynb
```diff
@@ -0,0 +1,893 @@
...
+    "    &= \\frac{1}{4}(\\braket{x}{x}+\\braket{x}{y}+\\braket{y}{x}+\\braket{y}{y}) \\\\\n",
```
frankharkins, line 83: Similarly for this. From this step it looks like |x + y> == |x> + |y>? It's not clear to me why that's the case.

frankharkins (commented):
Thanks for submitting this PR, without knowing much about variational algorithms, I am hoping this may be a nice introduction before students start on VQE / QAOA etc.

Can you add some information about the believed benefits of VQR over classical methods? Things like expected runtimes should be included here if possible. (If there is no benefit, that is also OK as long as we can frame the chapter as a valuable demonstration.)

I have also started reading through and will leave comments as I go.

content/ch-applications/Variational Quantum Regression.ipynb
clearya1: This is a good question as I should be more clear here. I am applying using the state $|\phi>$ after it has been acted on by the Hadamard gate on the first qubit. The |x + y> == |x> + |y> notation was a simplification, but I will get rid of it now as I see that it can be confusing. I have edited and recommitted. Hopefully it is a bit clearer now?

clearya1:
I did some very brief testing on run times using the simulators qiskit offers, and found that it scales linearly, like the classical version. I will put this result in and note that when run on an actual quantum computer we would expect to see a speed up of the run times (but perhaps not an improvement in the scalability).

I am somewhat new to the etiquette on Github though - once I commit changes to my fork, should I update my pull request somehow? 

Thank you for having a read through and commenting.

content/ch-applications/Variational Quantum Regression.ipynb
frankharkins: Yes this is much clearer, thank you

frankharkins:
@clearya1 can you please email me at Francis.Harkins@ibm.com or find me on qiskit.slack.com (Frank Harkins) to discuss further?

IvanIsCoding (commented):
@clearya1 I am reviewing this after the reading group organized by Frank

Firstly, this chapter could fill an interesting gap in the textbook. It is explicit in the text that there is no quantum advantage in the proposed VQR, and there is no issue in that. Indeed, the reviewing group quite liked it. For educational purposes, it would be good to have an algorithm that solves a "toy problem" that is as straightforward and focuses on introducing the core concepts of Variational Quantum Algorithms. We saw potential in using it as a warm-up chapter for VQE/VQLS later.

Secondly, there are good parts to this work. Picking linear regression as a problem is good because it is easy to introduce and can let us focus on the VQA concepts. The inner product using quantum computers was also great, the application is clever. It is the highlight of the chapter because it leads to the discussion of cost functions naturally. Also, the cost function is similar to the one used in VQLS, a bonus point for that.

Thirdly, there is a flaw in this work. I will highlight it to you. Currently, this flaw means the algorithm is not a VQA. It appears on this part:

```python
def calculate_cost_function(parameters):
    a, b = parameters
        
    yphi = a*xnorm/ynorm*inner_prod(y,x) + b*np.sqrt(N)/ynorm*inner_prod(y,ones)
    
    return (1-yphi)**2
```

The major issue is: `x` and `y` are constant. They're never updated after iterating the cost function. Hence, `inner_prod(y,x)` is also constant. There is no VQA. Consider this equivalent rewrite of the function as:

```python

MEASUREMENT_RESULTS_X = inner_prod(y,x)
MEASUREMENT_RESULTS_ONES = inner_prod(y,ones)

def calculate_cost_function(parameters):
    a, b = parameters
        
    yphi = a*xnorm/ynorm*MEASUREMENT_RESULTS_X + b*np.sqrt(N)/ynorm*MEASUREMENT_RESULTS_ONES
    
    return (1-yphi)**2
```

In practice, what happens is that we are doing classical optimization to find the optimal parameters `a` and `b` after executing the quantum circuit one. The algorithm is not a variational algorithm. It is classical post-processing.

The conclusion from the reading group is that the parameters should be in the circuit. Otherwise, it is not a VQA. The action item from the review is to parametrize the circuit. To do so, you may need to rewrite most of your code.

clearya1:
@IvanIsCoding Thank you very much for leaving these helpful comments and action item. I have fixed the implementation so that a different ansatz (parametrised by `a` and `b`) is fed to the quantum circuit at each iteration of the classical optimisation algorithm. This means that the quantum circuit is no longer just computing the same inner products `MEASUREMENT_RESULTS_X` and `MEASUREMENT_RESULTS_ONES` at every iteration. I also changed some of the text so that the ansatz is made more explicit, to prepare readers for ansatz's in later chapters. 

Let me know what you think!

PR #831, opened 2020-11-19
3.10 Fixed quality of 3-qubit grover circuit image

paniash:
# Changes made

The quality of the image used for the 3-qubit Grover circuit was poor. I generated the exact same image with LaTeX and the quality is much better.

**Before:**
pic-selected-201119-1820-21

**After:**
grover-tex


vabarbosa (approved):
thank you

PR #868, opened 2020-12-24
2.5 Suggestions and some minor fixes

amirebrahimi:
# Changes made
This was my first time reading through the chapter on universality, which I thoroughly enjoyed. I made some changes to the first paragraph and some of the last paragraphs.

# Justification
These are minor and the idea here is to leave something a little bit better than I found it.


amirebrahimi:
I noticed your changes, @frankharkins. Thank you. Anything else to do here?

PR #873, opened 2020-12-29
3.10 Fixed some typos and redundant code

luukament:
# Changes made + Justification

- Fixed formula of average tries needed in section 1 (under amplitude amplification). I think it reflected the worst case and not the average.
- Changed some text in the same section so that it corresponds better to the figures shown. The dashed line doesn't show in the first figure, but is introduced in the second figure, so I moved the mention of the dashed line down accordingly.
- Section 3.1 in the block "Details: Creating a General Diffuser", I propose we add a minus sign to make the formulas correct. It was a bit confusing to me to notice the missing minus sign in the formulas while that was only explained in the text afterwards.
- sudoku_oracle( ) took cbits as an argument but doesn't use it.


frankharkins:
Thank you!

Issue #885, opened 2021-01-16
Deprecation Warning due to the use of `QuantumCircuit.cu1` method.


oscar-defelice:
**Describe the bug**
In the current version of the textbook the `QuantumCircuit.cu1` method is often used when defining a generic $U(1)$ rotation.
However, the current version of Qiskit deprecates such a method and replace it with `QuantumCircuit.p`.
The bug is perfectly described by the warning
```
DeprecationWarning: The QuantumCircuit.cu1 method is deprecated as of 0.16.0. It will be removed no earlier than 3 months after the release date. You should use the QuantumCircuit.cp method instead, which acts identically.
  qc.cu1(theta,b,t)
```

**To Reproduce**

```python
from qiskit import QuantumCircuit
from math import pi

qc = QuantumCircuit(2)
qc.cu1(pi,0,1)
qc.draw()
```

**Version info***
```
{'qiskit-terra': '0.16.1',
 'qiskit-aer': '0.7.2',
 'qiskit-ignis': '0.5.1',
 'qiskit-ibmq-provider': '0.11.1',
 'qiskit-aqua': '0.8.1',
 'qiskit': '0.23.2'}
```

**My suggestion**
I think a misalignment between the current version of qiskit and textbook might be confusing. Thus, I would propose to replace the `cu1` methods with `p`.

paniash:
@oscar-defelice Thank you for pointing this out. I have made the changes in the relevant files and opened a [PR](https://github.com/qiskit-community/qiskit-textbook/pull/886) for the same. :-)

PR #886, opened 2021-01-16
Fix cu1 deprecation

paniash:
# Changes made

Changed `cu1` to `p` to avoid deprecation warnings, as suggested in #885.

# Justification

In the latest release of qiskit, `cu1` is replaced with `p` and is deprecated. To remove this warning when executing code, this PR has been made.


frankharkins:
@paniash thank you for your contribution, the `cu1` should be replaced by `cp`, not `p`. The `u1` gate is being deprecated in favour of `p`, and the `cu1` and `cp` are the controlled versions of those gates. I have fixed this in #887.

paniash:
Ah, my bad. No problem! :-)

PR #925, opened 2021-02-24
6.3 Correction in the order of the Lagrangian terms

jond01:
Corrected the terms in the Lagrangian

# Changes made
Kinematic term - (+), potential - (-)

# Justification
This is the definition of the Lagrangian


frankharkins (approved):
Spoke to @nbronn who approved this

jond01:
You are welcome :)

Actually, I spotted [other places](https://qiskit.org/textbook/ch-quantum-hardware/transmon-physics.html#4.-The-Quantized-Transmon-) that need to be corrected:
image
image

n and \phi should be hermitian operators, but from the definitions above they are anti-hermitian. Luckily, it does not change the following results.
I hope to make soon another PR that will address it.

jond01:
Oops, I accidentally closed this PR, reopening it for merging

Issue #888, opened 2021-01-22
3.3 Error in gate operation for message "11"


paniash:
**Describe the issue**
The gate operation `ZX` is applied when sending the classical message "11". In the table, the `X` gate is applied first, followed by the `Z` gate. However, when implementing the circuit in qiskit, the `Z` gate is applied first, followed by `X`.

**NOTE**: Both the operations yield the same state except a global phase of -1 differing between the two. This might make things confusing to the novice reader.

**Describe what you expect to see**

The "resulting state" for "11" as the message must be `|01> - |10>` instead (for it to be consistent with the implementation in qiskit).

frankharkins:
Thanks for your contribution. I'm closing this issue as this repository is being archived. You can find the latest version of the Qiskit Textbook online at <https://qiskit.org/learn/>, and the new content repository at [Qiskit/textbook](https://github.com/Qiskit/textbook). Please feel free to re-open this issue there if it's still relevant.

PR #931, opened 2021-03-02
3.11 Fixes issue #888

lebinyu:
# Changes made
1. Fixed calculation error in table 2 (superdense_table1.png), where the final state in the last row is |11>, without global phase.
2. Add subscripts to states in the table (superdense_table1.png & superdense_table2.png)  to make the state more clear for novice readers.
3. Fix typo in simulation (encode function). When sending message '11', X gate should be applied first before Z.

# Justification
1. -|11>+|10> -->H-gate on rightmost qubit--> -|1>(|0>-|1>)+|1>(|0>+|1>)=|11>


lebinyu:
#888

PR #938, opened 2021-03-09
Update quantum-key-distribution.ipynb

oscar-defelice:
`execute` import added

# Changes made

# Justification


frankharkins:
@oscar-defelice what is the justification for this? I should have removed all references to `execute` a while ago so it will be unused here.

frankharkins:
No worries! Thanks for keeping an eye out

PR #947, opened 2021-03-11
Fixed  broken cells output in section 'Running VQE on a Statevector S…

vabite:
…imulator'

# Changes made
Run the cells in order to correctly show the output results of the VQE simulation

# Justification
Before the cells were just showing error logs


frankharkins (approved):
Thank you!

PR #961, opened 2021-03-16
6.3 - transmon physics: corrected the definitions of \phi and n operators

jond01:
# Changes made
Now \phi and n are defined as Hermitian operators (with the ladder operators), previously they were defined as anti-Hermitian operators (both for QHO and transmon).
The final results are not affected by this change.

### Previously:
1.  image
2.  image

### Now:
1. image
2. image

# Justification
See also:
https://en.wikipedia.org/wiki/Quantum_harmonic_oscillator#Ladder_operator_method
Where in our case x->\phi, p->n


jond01:
Hi, any update on this?

nbronn:
Sorry for the delay, we have decided to keep sign conventions in line with [this Circuit Quantum Electrodynamics Review Paper](https://arxiv.org/abs/2005.12667), which is exactly the correction you have made. Thank you!

nbronn (approved):

frankharkins (approved):
Thanks!

Issue #960, opened 2021-03-16
Links are not working in Japanese translation website


kifumi:
**Describe the bug**
In Japanese version of website, the two icons of “Open in IBM Quantum Lab” and “Download as Jupyter Notebook” go to the error.

**To Reproduce**
1. Go to 'https://qiskit.org/textbook/ja/ch-algorithms/deutsch-jozsa.html'
2. Click on 'IBM Quantum Labを開く'
3. See error "404 : Not Found You are requesting a page that does not exist!"
and
1. Go to 'https://qiskit.org/textbook/ja/ch-algorithms/deutsch-jozsa.html'
2. Click on 'Jupyter Notebookのダウンロード'
3. See error "Page Not Found Sorry, but the page you were trying to view does not exist."

**Expected behavior**
Same behavior as English version of website.

**Version info***
device: Mac
OS: macOS Catalina
browser: Firefox, Chrome

**Screenshots**
![image](https://user-images.githubusercontent.com/19249007/111245636-4ba7fd80-8648-11eb-9645-1cd0bda9c573.png)

@vabarbosa , could you please take a look?

frankharkins:
Finally fixed, sorry for the wait

PR #963, opened 2021-03-16
Update _config.yml

vabarbosa:
@frankharkins @kifumi this should fix the issue with the broken download link in the Japanese pages

for the broken Quantum Lab link, we'll need to coordinate with the IQL lab team to update/sync the notebooks

related: #960

# Changes made

removed the localization directory from the Jekyll exclude

# Justification

removing localization directory from the Jekyll exclude allows its content to be copied and made available on the site for download


kifumi:
@vabarbosa thank you for fixing the broken download link.

> for the broken Quantum Lab link, we'll need to coordinate with the IQL lab team to update/sync the notebooks
who on the IQL lab team and how should I contact?

vabarbosa:
@kifumi that does not require anything from you. it is something @frankharkins coordinates with IQL. i believe he may be waiting on a different PR/update that will also require syncing the notebooks in IQL (and he will do them together).

Issue #954, opened 2021-03-13
Inconsistent labels for Bell state and the qubit to be sent in Teleportation (Ch. 3.10)


dhruvbhq:
**Describe the issue**
https://qiskit.org/textbook/ch-algorithms/teleportation.html
While going through the section on quantum teleportation, I observed some inconsistencies in the notations which might be potentially confusing for a novice reader:

1. In the early part of the section, from 'Overview' (https://qiskit.org/textbook/ch-algorithms/teleportation.html#overview) to 'Using the QASM simulator' (https://qiskit.org/textbook/ch-algorithms/teleportation.html#simulating-qs), $\ket{\psi}$ is used to denote the qubit that Alice wants to send to Bob via quantum teleportation.
However, in the following section 'Understanding Quantum Teleportation' (https://qiskit.org/textbook/ch-algorithms/teleportation.html#understanding-qt), there is an abrupt and unexplained change of notation. $\ket{\psi}$ is now being used to denote the shared Bell state of 2 qubits. The qubit to be sent is now denoted by $\ket{q}$. This is in disagreement with the notation being used till the point of 'Using the QASM simulator'.

2. With reference to the same sections, from 'Overview' to 'Using the QASM simulator', the probability amplitudes for the state to be transmitted (ie $\ket{\psi}$) are taken to be $\alpha$ and $\beta$. However, in 'Understanding Quantum Teleportation', the probability amplitudes are now given by $a$ and $b$ for the qubit to be sent. To make things potentially even more convoluted for a beginner, the function create_bell_pair() in an intervening section (https://qiskit.org/textbook/ch-algorithms/teleportation.html#how) uses 'a' and 'b' as arguments for qubits to initialize in a Bell state.

**Describe what you expect to see**
It might be a good idea to keep the notation in the section 'Understanding Quantum Teleportation' (https://qiskit.org/textbook/ch-algorithms/teleportation.html#understanding-qt) consistent with the preceding matter. The following could be possible solutions to the above respectively:

1.  In 'Understanding Quantum Teleportation' , the qubit to be sent should be labelled by $\psi$ and the Bell state should use a letter other than $\psi$.
2. In 'Understanding Quantum Teleportation', the probability amplitudes for the state to be transmitted (ie $\ket{\psi}$) should be taken as $\alpha$ and $\beta$ in the ensuing explanation, in place of $a$ and $b$.

dhruvbhq:
[...]

PR #971, opened 2021-03-21
fixed issue 954 by replacing state labels and coefficients in markdow…

dhruvbhq:
3.10, Fixed issue #954 

# Changes made

In the section 4. (Understanding Quantum Teleportation) of 3.10, did the following changes in the markdown cells, to make the explanation consistent with text appearing earlier in the same notebook page:
1. Changed the shared entangled state's label from $$|\psi>$$ to $$|e>$$
2. Changed the label for the qubit to be sent by Alice from $$|q>$$ to $$\psi$$
3. Changed the probability amplitudes $$a$$ and $$b$$ to $$\alpha$$ and $$\beta$$ respectively.
4. Replaced the graphic teleportation-transformation.png with a table created via latex array environment.
# Justification

1. Changes 1. and 2. were required because text matter before '4. Understanding Quantum Teleportation' uses $$|\psi>$$ for the qubit to be sent by Alice. However, in '4. Understanding Quantum Teleportation', there was an abrupt and unexplained change of notation to $$|\psi>$$ being used for the shared entangled state and $$|q>$$ being used for qubit to be sent. This was creating an inconsistent notation.
See the following snippets of the 2 subsections in 3.10 in their existing form:
image

image
2. There was a similar inconsistency in letters being used to denote the probability amplitudes for the qubit to be sent by Alice. Earlier part of 3.10 uses $$\alpha$$ and $$\beta$$, while '4. Understanding Quantum Teleportation' uses the notation $$a$$ and $$b$$. This justifies changes 3. and 4. 
(Pl. refer to the attached snapshots).
3. Further explanation for change 4.: Earlier, there was a graphic teleportation-transformation.png in one markdown cell in '4. Understanding Quantum Teleportation' in 3.10. This png file contained the inconsistent notation $$a$$ and $$b$$. In order to use $$\alpha$$ and $$\beta$$, this png file has been replaced by a table in latex created via array environment. It conveys the same information as the png file, while care has been taken to replace $$a$$ and $$b$$ with $$\alpha$$ and $$\beta$$ respectively.

### DIFF
--- a/content/ch-algorithms/teleportation.ipynb
+++ b/content/ch-algorithms/teleportation.ipynb
@@ -22383,19 +22383,19 @@
     "\n",
     "#### Step 1\n",
     "\n",
-    "Quantum Teleportation begins with the fact that Alice needs to transmit $|q\\rangle = a|0\\rangle + b|1\\rangle$ (a random qubit) to Bob. She doesn't know the state of the qubit. For this, Alice and Bob take the help of a third party (Telamon). Telamon prepares a pair of entangled qubits for Alice and Bob. The entangled qubits could be written in Dirac Notation as:\n",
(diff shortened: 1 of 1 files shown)


frankharkins (approved):
Thanks!

PR #981, opened 2021-03-24
Argument `n` explicitly passed to function `get_initial_layout`

vabite:
5.1 Introduction to Quantum Error Correction using Repetition Codes

# Changes made
1. Argument `n` explicitly passed to function `get_initial_layout`
2. Fixed bug in accessing the list of circuits to be executed
3. corrected typos

# Justification
1. Before it was taken from the global scope, thus the function did not run outside the for-cycle
2. The notebook crashed. The code seemed to have no problems at run time simply because the access to the bugged piece of code was not enabled by a `step_2` variable arbitrarily set to `False`
3. no need to explain


frankharkins:
Merged in #987

PR #982, opened 2021-03-24
update position of language selector

### LINKED ISSUE #978: Move the language switch to a more visible position
**Describe the problem**
Many people might not be unaware of the language switch at its current location.
image

**Describe the expected solution**
Can you move the switch to a more visible position, for example at the top of the page?  In the Qiskit Documentation, it was also at the bottom of the page, but recently moved to the top.
image

vabarbosa:
# Changes made

the language switcher position has been changed to fixed so it always visible as the sidebar scrolls

image

# Justification

Resolves #978 

making the switcher position fixed and always visible ensure users are aware of the ability to change language

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -172,13 +172,15 @@ Textbook numbering starts in Chapter 0.
       {% endunless %}
     {% endfor %}
   </ul>
-  <p class="sidebar_footer">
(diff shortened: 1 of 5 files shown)


frankharkins (approved):

frankharkins:
Thank you!

PR #987, opened 2021-03-25
Typo: than -> that

vabite:
5.1 Introduction to Quantum Error Correction using Repetition Codes
5.2 Measurement Error Mitigation

# Changes made
1. Passed argument `n` to function `get_initial_layout`
2. Fixed bug in accessing the list of circuits to be executed
3. Fixed typos

# Justification
1. the function `get_initial_layout` took argument `n` from the global scope, thus was not usable outside of the for cycle
2. a bug was present, and the notebook didn't crash just because an auxiliary variable `step_2` was set to `False` disabling the access to the bugged piece of code
3. no need to explain


frankharkins (approved):

frankharkins:
Thank you!

PR #988, opened 2021-03-26
Update deutsch-jozsa.ipynb

3abc:
1.3 4. Fix a typo. Should be 1/√2ⁿ instead of 1/2ⁿ.

# Changes made

# Justification


paniash (commented):
There are two summation notations, each corresponds to a `\frac{1}{\sqrt{2^n}}`, thus making it `\frac{1}{2^n}`.

3abc:
> There are two summation notations, each corresponding to a `\frac{1}{\sqrt{2^n}}`, thus making it `\frac{1}{2^n}`.

You are right. Sorry.

PR #992, opened 2021-03-27
3.2 Fixed math typo

paniash:
# Changes made

Under section 1.3, the probability of measuring `|0>^n` has its coefficient as `1/2^n` within the modulus squared when it really should be `1/\sqrt{2^n}`.

# Justification

For a collection of n qubits, probability of measuring a particular eigenstate is the modulus square of `1/\sqrt{2^n} x <relevant_summation_terms>`.


frankharkins (approved):
Thanks!

PR #1000, opened 2021-03-30
Fix for notation inconsistency

### LINKED ISSUE #994: A typo in 3.9 Quantum Counting
**Describe the issue**
About the error calculation, there is a sign typo that is inconsistent with the textbook Nielsen and Chuang:
In the code:
err = (math.sqrt(2*M*N) + N/(2**(m-1)))*(2**(-m))
According to Nielsen and Chunag, it should be
err = (math.sqrt(2*M*N) + N/(2**(m+1)))*(2**(-m))

The same typo also occurs at the function calculate_M.

paniash:
# Changes made

Fixed #994. Changed '-' to '+' and appropriately ran the rest of the cells.


content/ch-algorithms/quantum-counting.ipynb
```diff
@@ -305,7 +305,16 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 28,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%config InlineBackend.figure_format = 'svg'"
```
frankharkins, line 312: Thanks for rendering as SVG! Looks much better on the web. Can you please remove this line though? It should be set in the config and not in the notebook (to keep things simpler for readers).

paniash: Ah, I forgot to remove it. Thanks for pointing out!

frankharkins:
Thank you!

PR #951, opened 2021-03-12
9, Adding CHSH demo notebook

adcorcol:
# Changes made
Created a notebook outlining and demonstrating the violation of the CHSH inequality

# Justification
This is an interesting addition to the Demos chapter

PR #1021, opened 2021-04-04
Ch9, Minor edits to CHSH demo

adcorcol:
# Changes made
Simplified the math pertaining the CHSH bound after a discussion with Frank. 
# Justification
This was as a result of a discussion with Frank


frankharkins:
Closing as merged in #951

PR #1024, opened 2021-04-05
Fix case-sensitivity of matplotlib spelling

paniash:
# Changes made

Changed "MatPlotLib" to "Matplotlib"

# Justification

As stated in the official [bibtex](https://matplotlib.org/stable/citing.html) entry, it is to be spelled as "Matplotlib".


frankharkins (approved):
Thanks

PR #1040, opened 2021-04-15
2.2 Fixed inconsistent scaling in Dirac notation

jtweigle:
An expression in section 2.2 looks like:
$\left\langle \psi_0 |U^\dagger U| \psi_1 \right\rangle$

Whereas it should look like:
$\left\langle \psi_0 \right|U^\dagger U\left| \psi_1 \right\rangle$

# Changes made
Added \left and \right where they were missing.

# Justification

The expression is much easier to read if $\left\langle \psi_0 \right|$ looks like one item, and $\left| \psi_1 \right\rangle$ looks like one item.


frankharkins (approved):
Thanks!

Issue #1051, opened 2021-04-29
[1.4] Issue with the X-basis measurement 


declanmillar:
**Describe the issue**

<img width="849" alt="Screen Shot 2021-04-29 at 10 30 42" src="https://user-images.githubusercontent.com/14238050/116531149-c49ca500-a8d6-11eb-8cfc-6ad5cf04b6b2.png">

In Section 1.4 (see attached image), the paragraph on the X-measurement seems inconsistent with the code example above (see attached image). Also, as it's using the `qasm_simulator`, I don't think it's "outputting" the state vector.

**Describe what you expect to see**

I would expect that either the code example was tweaked to match the paragraph, or that the paragraph was tweaked to match the code example :).

paniash:
@declanmillar Hi! In the example, the intention is to *measure* the qubit along the X axis and hence using `qasm_simulator` makes sense.  :-)

declanmillar:
Hi @paniash :). Yep, very true, but the comment on the last line of the code says "# Display the output state vector", which I think isn't quite right, given the sim in use. That's just a small comment from me though, I think the broader issue (if I've understood the intention) is that the paragraph the below plot doesn't match what the code is doing.

paniash:
@declanmillar Yes, the comment could be changed to better make sense of what's being actually done.

Regarding the broader issue, you're right. The end result should always be 1 when the initial state is |-> before running `x_measurement()`. :-)

paniash:
@declanmillar I've made the changes as a PR. Please let me know if it's alright or we could discuss over on #1054.

PR #1054, opened 2021-04-30
Ch 1.4 Fixed paragraph on X-measurement

paniash:
# Justification

The measurement along X-basis is done with the initial state being |->. Hence, the measurement will always yield |1> and not different results every time the measurement is done. Change is done to accommodate this disparity. See issue #1051.


frankharkins (approved):
Thank you!

PR #1055, opened 2021-04-30
4.2.2 changed qc to qc_image in NEQR example

gines-carrascal:
# Changes made

changed qc to qc_image and t_qc to t_qc_image in NEQR example

# Justification

variables user refers to objets already created in previous section, not to the circuit of this section (posibly a copy and paste error)


frankharkins (approved):
Good spot, thank you!

PR #1057, opened 2021-05-01
2.3 Fix typo introduced in previous commit ccc3ff4

sikfeng:
# Changes made

Removed extra "if the" from quick exercise question.
# Justification

image

Its a typo.


frankharkins (approved):
Thanks

Issue #970, opened 2021-03-20
[3.5] Inconsistency with the notation in chapter 3.5


byulen:
**Describe the issue**
There is an inconsistency with the notation in chapter 3.5 in particular between section 1 (introduction) 
and section 4 (The Quantum Fourier Transform).

In section 1, |x> state is not defined at all. Actually, I think |x> is the same thing with |i> mentioned in the expression sum_{i=0}^{N-1} x_i |i>, that is, it is one of the basis states. Here, the vectors (x_0,x_1,...,x_{N-1}) and (y_0,y_1,...,y_{N-1}) are mentioned. The values of these coefficents are complex numbers. In section 4, |x> and |y> states are defined as |x_1 x_2 x_3 ... x_N> and |y_1 y_2 y_3 ... y_n> where now y_1, y_2, y_3 etc. are bit values taking values 0 and 1.  

Summarizing, the vector (x_0,x_1,...,x_{N-1}) in section 1 should be clearly distinguishable from the state |x_1 x_2 x_3 ... x_N> in section 4.

**Describe what you expect to see**

In order to remove any ambiguity, I recommend to use different notations for probability amplitudes and basis states. It would be much better if the notation is consistent everywhere in chapter 3.5.

I suggest to use |X> =  (x_0,x_1,...,x_{N-1}) = sum_{i=0}^{N-1} x_i |i> for the vector that carries the coefficients to be Fourier transformed. Similarly, 
|Y> = (y_0,y_1,...,y_{N-1}) =sum_{i=0}^{N-1} y_i |i> in section 1. Also, I think using the notation |k> and |j> instead of |x> and |y> will be more appropriate.

Note that now |Y> = U_{QFT} |X> where U_{QFT} = 1/sqrt{N} sum_j sum_k w_N^{jk} |j> <k| can be written without ambiguity with the new notation.

singhmeet11:
can I work on this ? @frankharkins

PR #1058, opened 2021-05-01
(chapter 3.5)Fixed the inconsistency issue in QFT

singhmeet11:
The issue resolved is given in #970 

[3.5] Fixed the inconsistency in QFT chapter's

# Changes made
 The main aim was to eradicate the confusion between the |x> in section 1 and |x> in section 4. I have made the following changes:

1. |x> was not defined clearly, I defined a |X>(so that people don't get confused again) , I also changed the respective index from i to j (for X) and from i to k(for Y). 
2.  The mapping from |x> to |y> (which seemed confusing to me, because it nowhere mentions that we are talking about the whole vector or just one element), I used |j> and |k> for making this clearer .
3. The last change was in the U_{QFT} operator where I changed |x> and |y> to |j> and |k>.

No changes have been made anywhere except section 1. 

 

# Justification

1. This resolves the issue of intermixing of notations in section 1 and section 4. As there is a clear distinction between |X> and |x>.
2. Further replacing x and y by j and k in the powers of e, at a few places makes it easier to interpret the equations.
3. The overall linear algebra fits much better


divshacker (changes requested):
Hey @singhmeet11  you have just added two extra files which are not required. You have to just edit the quantum-fourier-transform.ipynb file and just make a PR which you have done already. but you need to delete these extra files you have added in your branch

singhmeet11:
@divshacker thanks for helping me out. Does this look good ?? 
And sorry if its still wrong a little new to all this

divshacker (dismissed):
Now it's looks good.

content/ch-algorithms/quantum-fourier-transform.ipynb
```diff
@@ -55,7 +55,7 @@
     "\n",
     "where $\\omega_N^{jk} = e^{2\\pi i \\frac{jk}{N}}$.\n",
     "\n",
-    "Similarly, the quantum Fourier transform acts on a quantum state $\\sum_{i=0}^{N-1} x_i \\vert i \\rangle$ and maps it to the quantum state $\\sum_{i=0}^{N-1} y_i \\vert i \\rangle$ according to the formula\n",
+    "Similarly, the quantum Fourier transform acts on a quantum state $\\vert X\\rangle =\\vert x_0  x_1  ...  x_{N-1}\\rangle= \\sum_{i=0}^{N-1} x_j \\vert j \\rangle$ and maps it to the quantum state $\\vert Y\\rangle = \\vert y_0 y_1 ... y_{N-1}\\rangle = \\sum_{i=0}^{N-1} y_k \\vert k \\rangle$ according to the formula\n",
```
frankharkins, line 58:
```suggestion
    "Similarly, the quantum Fourier transform acts on a quantum state $\\vert X\\rangle = \\sum_{j=0}^{N-1} x_j \\vert j \\rangle$ and maps it to the quantum state $\\vert Y\\rangle = \\sum_{k=0}^{N-1} y_k \\vert k \\rangle$ according to the formula\n",
```

content/ch-algorithms/quantum-fourier-transform.ipynb
```diff
@@ -66,14 +66,14 @@
     "This can also be expressed as the map:\n",
     "\n",
     "\n",
-    "$$\\vert x \\rangle \\mapsto \\frac{1}{\\sqrt{N}}\\sum_{y=0}^{N-1}\\omega_N^{xy} \\vert y \\rangle$$\n",
+    "$$\\vert j \\rangle \\mapsto \\frac{1}{\\sqrt{N}}\\sum_{y=0}^{N-1}\\omega_N^{jk} \\vert k \\rangle$$\n",
```
frankharkins, line 69:
```suggestion
    "$$\\vert j \\rangle \\mapsto \\frac{1}{\\sqrt{N}}\\sum_{k=0}^{N-1}\\omega_N^{jk} \\vert k \\rangle$$\n",
```

frankharkins (changes requested):
Thanks for having a go at this! I've suggested some changes:
- I think the `\\vert x_0  x_1  ...  x_{N-1}\\rangle` is maybe misleading as we generally use the label of a ket to refer to it's computational basis state, whereas here it seems `x_i` is the amplitude of the ith computational basis state.
- There are a few places I think the letters don't match up, I've suggested changes that I think makes the equations look right.

these equations are very dense though so it's possible I'm mistaken. Let me know what you think. 

Thanks!

content/ch-algorithms/quantum-fourier-transform.ipynb
frankharkins, line 58:
```suggestion
    "Similarly, the quantum Fourier transform acts on a quantum state $\\vert X\\rangle = \\sum_{j=0}^{N-1} x_j \\vert j \\rangle$ and maps it to the quantum state $\\vert Y\\rangle = \\sum_{k=0}^{N-1} y_k \\vert k \\rangle$ according to the formula\n",
```

frankharkins (approved):
Thanks!

Issue #934, opened 2021-03-04
Pulse code example updates (Qiskit Advocate Project)


nkanazawa1989:
Currently some pulse code examples are written in the old programming model based on `Schedule`. I'll update these code examples with the _pulse builder_ syntax which is becoming a new standard of Qiskit Pulse programming. This work will be done as a part of the qiskit advocate project, with the contribution of the team @faraimazh @kUmezawa @0sophy1 @splch.

See below for details:
https://github.com/qiskit-community/qiskit-advocate-mentorship-program/issues/2

0sophy1:
so excited to join!

faraimazh:
Let's hack on!

kUmezawa:
Hello !

aasfaw:
Well done Naoki and team. Good luck!

splch:
You can't beat🎶 Pulse!🎵

nkanazawa1989:
Thanks @frankharkins for assignment :)

frankharkins:
Thanks everyone for your great work!

PR #1059, opened 2021-05-02
6.1 Changed to Builder syntax

kUmezawa:
# Changes made

Changed how to write pulse schedule to Builder syntax

# Justification


nkanazawa1989 (changes requested):
Great job! @kUmezawa. This section seems to be correctly updated with the builder syntax. I have two major update suggestions.

1. Use call instruction for `pi_pulse` and `x90_pulse`.

In discriminator and T1/T2 experiments, you define the same Gaussian pulse in every cell. This makes the example codes longer, and might be a burden for learners. For example, you can define
```python
with build(backend) as pi_pulse:
    play(Gaussian(drive_duration, pi_amp, drive_sigma, 'pi_pulse'), pulse.drive_channel(qubit))
```
in the new cell before entering into 3.2 (right after we calibrated `pi_amp`) as a result of calibration. Then, in the following cells, you can `call(pi_pulse)` instead of redefining the instruction.

2. Use population in T1/T2 experiments
In the cell 33 you calibrated a discriminator function, but this is not used in the following experiment. Usually, the result of T1/T2 experiment is much easier to see in the population as a vertical axis, rather than in the signal. To do this, you can request `meas_return='single'` and
```python
iq_data = result.get_memory(i)[:, mem_ind] * scale_factor
population = sum(map(classify, data)) / shots
```

In the DD result, the population seems to be rapidly decreasing. Since the code seems ok, perhaps this is due to poor calibration of pi pulse or TLS. What happens if you rerun this experiment? Because T2 value is actually almost double of T1, the DD doesn't help so much to improve the coherence time of this specific qubit. However this should not deteriorate coherence value.

You see some higher level excitation in the result of 3.2, perhaps the pi pulse is poorly calibrated without DRAG. This yields the second level excitation as pi pulse is repeated in the sequence. Can you try rerunning the notebook? If this doesn't fix the problem, maybe XY-4 would help. Note that the chip version is now `'2.4.6'`, this means the armonk processor is replaced recently. If the hardware is improved and the charge noise is suppressed, small miscalibration of echos tends to reduce coherence time.

Minor comments:

- In cell 8, there are many `arg1,arg2`. This violates PEP8 rule. You need to insert space between args like `arg1, arg2`.

- There are many inactive lines. For example, `# print(job.job_id())` in cell 11. Please remove them if you don't use.

- In cell 18, `measure` in the comment should be `mem_slot` in new syntax (we are no longer define measure schedule).

- In cell 32, please remove `figsize=[4,4]` and add `plt.axis('square')` instead? Because the distribution of scatters are not equal in the x and y axis (you have 0 blob and 1 blob in the same plot), setting the aspect ratio = 1 may distort the appearance of blobs. 

Notes:
This is a part of #934

lcapelluto:
Very cool! This is awesome; well done.
I especially like the statements such as `rabi_schedules = [rabi_sched.assign_parameters({drive_amp: a}, inplace=False) for a in drive_amps]`. We can't use this syntax for the T1 experiment because the latest terra code hasn't been released?

If Naoki's comments are addressed, I would approve it.

Small note: Does the builder take `alignment` as an initialization param, like `pulse.builder(..., alignment=...)? Sequential alignment seems to be the only one that is used for most (or all) of these programs, so it would reduce the extra indentation if we could just set it from the start. I'm also fine with it if we leave it as is.

nkanazawa1989:
Thanks Lauren :) Yes, with the next release of terra we can parametrize T1 experiment as well (I hope @kUmezawa will continue after next release). Thanks again for the comment about initial alignment. Specifying `alignment='sequential'` will enable you to remove the alignment context within the builder.

frankharkins:
Thank you so much! Let me know when this is ready to be merged

kUmezawa:
@nkanazawa1989 
I have corrected the content you pointed out. Please check this. Thank you.

nkanazawa1989:
This is awesome! It's almost good to go. Can you update below minor errors? Then we can merge this PR.

Cell18, 19

There is an indent error.
```python
    pulse.play(pulse.Gaussian(duration=drive_duration,
                                 sigma=drive_sigma,
                                 amp=drive_amp,
                                 name='freq_sweep_excitation_pulse'), drive_chan)
```
should be
```python
    pulse.play(pulse.Gaussian(duration=drive_duration,
                              sigma=drive_sigma,
                              amp=drive_amp,
                              name='freq_sweep_excitation_pulse'), drive_chan)
```

Cell 28

The comment seems to be mistake. I cannot find 2A. Please remove this comment.
```
# We found this in Part 2A above
```

Cell 59
```python
 if _!=num_sequence-1:
```

Usually `_` indicates a throw away variable in for loop.
https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python

Please use some meaningful variable name such as `loop_counts`. Also you need a space around variable:
```python
 if loop_counts != num_sequence-1:
```

Cell 63

Please remove dot after 4. `4.` indicates a floating value of 4.0, but this is number of delays per unit XY4 sequence thus should be integer. Just a nitpick.
```
4.*num_sequence*taus_sec -> 4*num_sequence*taus_sec
```

kUmezawa:
@nkanazawa1989 
Thank you for your review. I update minor errors.

nkanazawa1989 (approved):
Thanks a lot for your great work. This chapter becomes very good example to learn the feature of pulse builder! @kUmezawa

frankharkins (approved):
Amazing, thank you!

frankharkins:
@kUmezawa can you contact me on Slack please?

PR #1063, opened 2021-05-04
Update single-qubit-gates.ipynb

suraniakbar:
1.4 Single Qubit Gates
2. Digression: The X, Y & Z-Bases'
'Quick Exercises

# Changes made
In '2. Digression: The X, Y & Z-Bases', under 'Quick Exercises', I replaced 'eigenvalues' with 'eigenstates' in the following line: 
3. Why would we not see these eigenvalues appear on the Bloch sphere?

The new line should be:
3. Why would we not see these eigenstates appear on the Bloch sphere?

# Justification
Eigenstates, not eigenvalues, can be plotted on Bloch sphere.

- Akbar Surani


content/ch-states/single-qubit-gates.ipynb
```diff
@@ -4502,7 +4502,7 @@
     "### Quick Exercises\n",
     "1. Verify that $|+\\rangle$ and $|-\\rangle$ are in fact eigenstates of the X-gate.\n",
     "2. What eigenvalues do they have? \n",
-    "3. Why would we not see these eigenvalues appear on the Bloch sphere?\n",
+    "3. Why would we not see these eigenstates appear on the Bloch sphere?\n",
```
frankharkins, line 4505:
```suggestion
```

content/ch-states/single-qubit-gates.ipynb
```diff
@@ -4502,7 +4502,7 @@
...
     "4. Find the eigenstates of the Y-gate, and their co-ordinates on the Bloch sphere.\n",
```
frankharkins, line 4506:
```suggestion
    "3. Find the eigenstates of the Y-gate, and their co-ordinates on the Bloch sphere.\n",
```

frankharkins (changes requested):
Thanks for making a contribution! I don't think your change is correct as you can show the |+> and |-> states on the bloch sphere, and these would be the eigenstates the question is referring to. 

I see the confusion as the original question is phrased poorly. I think it's trying to get the reader to think about global phase but really it's just unclear and doesn't have a "right" answer. I instead suggest removing it.

frankharkins:
No reply so I've committed the suggested changes. If you see this and disagree please open an issue or make another PR.

PR #1067, opened 2021-05-06
Fix issue 1061

rafal-pracht:
# Changes made

# Justification


diemilio:
Hi @rafal-pracht. By looking at the changes you implemented, I can see you fixed the problem qc.measure() problem in 2 parts of the notebook, but please note that there are 3 places where it needs to be corrected.

rafal-pracht:
Thanks @diemilio that you note this. This has no impact on the histogram therefore I missed it.
I've just correct this.

PR #1086, opened 2021-05-14
4.1. Update the HHL tutorial to account for the changes after Qiskit's refactoring

anedumla:
# Changes made
Updated and changed completely section 4.A. to explain how to use the new `linear_systems` package in Qiskit terra (previously the HHL implementation lived in Aqua and was completely different).

# Justification
The new implementation is completely different than the former, which is the one currently shown in the textbook.


hykavitha:
@frankharkins

hykavitha:
@frankharkins

frankharkins:
Thanks, @hykavitha will help review this

hykavitha:
It's really very nice of you to use `TridiagonalToeplitz` class & device the hhl tutorial. This is much clearer than the hhl_tutorial which is there now. Thank you so much for this. Really appreciate the extra explanation. 

Here are some suggestions to consider:

1.  Change the variable name from `quantum_solution` to something like `naive_hhl_solution` & any texts that refers to this variable. 
2. Could you please explain What the Euclidean norm is ? and Why that’s important here?
3. Is there an other way we can verify the quantum circuit result to classical result? perhaps use state-vector simulator and compare the simulated states? It might not be clear that euclidean norm is the way to verify classical vs quantum?

anedumla:
@hykavitha Thanks a lot for your input, I have updated the notebook to incorporating your suggestions. Hope is now all clearer :)

hykavitha:
@frankharkins : i'll take the code to merge. it sounds good after going through.

frankharkins (approved):
Thank you!

PR #1104, opened 2021-05-20
Fixed lookuptable function argument order

### LINKED ISSUE #1101: Section 5.1 (Quantum Error Correction) possible swapped arguments in code example using function "lookuptable_decoding"
**Describe the issue**

In the following textbook section:

https://qiskit.org/textbook/ch-quantum-hardware/error-correction-repetition-code.html#Lookup-table-decoding

there is a code example at the end of that section (just before the last paragraph) which calls the function _lookuptable_decoding()_  as follows:

`P = lookuptable_decoding(raw_results,table_results)  `
`print('P =',P)
`

If one compares the order of the two arguments _(raw_results,table_results)_ to the description in the actual function:

`def lookuptable_decoding(training_results, real_results):`

located in the file: 

https://github.com/Qiskit/qiskit-ignis/blob/master/qiskit/ignis/verification/topological_codes/fitters.py

it appears that the order of the arguments as written in the textbook code example might be reversed and lead to possible confusion by the reader.

I believe that _table_results_ should come before _raw_results_ in the function argument call sequence as written in the textbook code example, to be consistent with the description in the textbook and the actual function behavior.

divshacker:
# Changes made

fixes #1101 
# Justification

All details can be found in the above liked issue.


frankharkins (approved):
Thank you!

divshacker:
I don't why it is merging automatically. XD

PR #1129, opened 2021-05-29
Ch 3.3 Updated svg figures to LaTeX

paniash:
# Changes made

The svg figures for the quantum circuits were pretty small and were hard to read on a smaller screen. I generated the exact same circuit figures in LaTeX and saved them as png files.


frankharkins (approved):
Thanks!

PR #1135, opened 2021-06-02
6.7 Update pulsyntax

0sophy1:
Updated with new pulse syntax and code simplified.

Part of Qiskit Advocate mentorship program.

# Changes made
Updated old syntax especially try to avoid using instmap. 
# Justification

Pulse syntax updated.
Currently, minor bugs make it difficult to use PulseSimulator, but I've also provided a guide on it.


nkanazawa1989:
This is a part #934

nkanazawa1989:
Thanks @0sophy1 ! This drastically improves readability of example code, leveraged by the pulse gate. I really like new version. First of all, this PR chapter will have two future updates.
- remove `!pip install qiskit-aer==0.7` once the pulse simulator bug is fixed and the patch is merged into stable.
- use parameter feature for parameter scan (this will further simplify the example code) after next Qiskit terra is released.

I'm not sure if you can continue to take care of this chapter @0sophy1 but someone should be responsible for these updates.

### Review comments

- cell 153
```
cr_pulse = inst_sched_map.get('cx', (qc, qt)).filter(channels = [ControlChannel(1)], instruction_types=[Play]).instructions[qc][qt].pulse
```
Instruction tuple indicates `(t0, instruction data)`, not qubits. This should be `[0][1]`. This means taking instruction data from the first instruction tuple. The schedule is filtered by control channel and play instruction, and thus the first instruction in the list should be `Play`ing cross resonance pulse.

Also you don't need to keep comment block.
```
"""Retrieve the cross resonance pulse channel and waveform
for the qubit pair from the cx (CNOT) schedule.
    
Args:
    qc: control qubit index
    qt: target qubit index
"""
```
This is the format of docstring for functions or class methods. If you want to add some comments, you can add _Retrieve the cross resonance pulse..._ to the markdown cell just above the code.

- cell 156
The comment `# middle sample will contain the amplitude` doesn't make sense. I think you can remove this.

- md cell comment just above cell 158
_We keep this in mind because this particular cr_dur will only..._ `cr_dur` is no longer defined. You can replace this with "duration of cross resonance pulse".

- cell 158
You don't need to call `transpile` and `schedule` for each single entry. These functions take a list of entries. So you can
```python
    tomo_circs = []
    for width in cr_times:
        ...
        tomo_circs.append(tomo_circ)

    tomo_circs_transpiled = transpile(tomo_circs, backend, optimization_level=1)

return schedule(tomo_circs_transpiled, backend)
```

- cell 169
you defined `qc` and `qt` again (they are already defined)

- cell 184
`tomo_cr_params` doesn't make sense in this context. Perhaps `ramsey_cr_params`?

sdg-h is not X90 pulse. Actually this sequence if used to project qubit in Y measurement basis, and not often seen in the context of Ramsey experiment. Usually you need to use `ramsey_circ.sx` (this changes output -- it starts from <Z> = -1 at t=0, so you need to rerun the ramsey experiment).
```python
        #apply X90 pulse
        ramsey_circ.sdg(qc)
        ramsey_circ.h(qc)
```

I can approve this PR after those points are addressed. Any other comment about pulse gate part? @lcapelluto

0sophy1:
Thanks @nkanazawa1989! I need to be more careful before PR. I've fixed  (and also fixed the pulse.offset to pulse.phase_offset) and please check this again.

lcapelluto:
Very cool! Pulse gate usage looks good to me!

Some feedback:

In the introduction you have a typo: `so for know let us focus on the Hamiltonian` -> `know` should be _`now`_.

I don't think cell 3 would make sense to anyone who hasn't seen it before, so you might want to add a link to the other docs or try to make your implicit assumptions explicit. For example:
 -  "First we grab all the pulse-level gate calibrations, then we get the CNOT operation between two qubits. We already know from experience with this particular backend that the CNOT operation contains a CR pulse implemented as a `GaussianSquare` on a particular channel, `ControlChannel(1)`. We use this knowledge we have about the form of that instruction to extract it from the overall CNOT pulse.Schedule by using the `filter` method."

I don't understand what you are doing with cells 4 and 5. Seems like an odd way to copy a pulse. If the CR pulse was actually a `Waveform` (in sample form), `parameters` should be empty... so I don't see what the difference is between `cr_pulse` and `my_cr_pulse`.

Cell 7, comment `#21` is not helpful.

For defined functions, I love that you included docstrings. I think it would be valuable to document what is returned, too!

Fit seemed to fail for one of the <X(t)> plots

nkanazawa1989:
Thanks Lauren! Comparing two CR pulses in the same plot (by changing channel) is from the original notebook. I agree this is bit weird since pulse is tightly coupled to channel. Perhaps we can remove cell4 and 5? Now it is obvious that we get the same waveform since it copied the entire parameters from original pulse (previously it assumed risefall parameter instead of copying `sigma` from original pulse - I don't know why). 

> If the CR pulse was actually a Waveform (in sample form), parameters should be empty

Good point. In this case we need to assume `sigma`, `amp` and `risefall` to continue the experiment (then we need comparison like cell4 and 5). I think adding
```
assert isinstance(cr_pulse, GaussianSquare)
```
to cell 3 should be fine.

nkanazawa1989:
Perhaps we need to tune the initial parameter `p0` in cell16 to improve fit of <X(t)> plot?

0sophy1:
@lcapelluto  Thank you for your kind review. About the cell 4 and 5, I thought it was good to remain because the two way of build CR pulse seems helpful in some sense. I've slightly modified my_cr_pulse declaration part and now checking the <X(t)> parameter part. according to that, X(t) has a relatively small axis scale so I've plot again with the same axis with others. Screen Shot 2021-06-03 at 11 56 28 AM It still seems like slightly misfitted but the calculated parameter value is similar with the previous textbook one.

nkanazawa1989:
Hi @0sophy1 , thanks for updating the notebook.

I realized that you use `sdg-h` to create Ramsey, but this results in bit weird sequence,

image

namely you drive X90 pulse to prepare the target qubit (ignoring second phase since this is just frame change, i.e. frame change just updates the phase of _following_ pulses), and apply Y90 pulse before measurement (second pulse is phase shifted by pi/2 because of VZ of the first pulse). This sequence is somewhat like (imperfect) XY dynamical decoupling, and unintentionally using such sequence may confuse experimentalists (this gives you phase shifted Ramsey oscillation by pi/2).

If you don't have a special reason to use this, please replace `sdg-h` with `sx` (this will give you standard Ramsey experiment sequence).

Regarding the fitting, you updated the initial value p0 but seems like still you cannot get the perfect fitting. I agree this experiment usually give us perfect fit, because of risefall part of GaussianSquare is not taken into account to the fit model. In the fit result there seems to be finite phase mismatch (i.e. you can also see error in <Y> plot in the time [0, 1000]). Perhaps you need to use global optimization type fitter such as dual annealing. But this small error can be ignored in the textbook. How do you think @frankharkins ?

0sophy1:
@nkanazawa1989 Thank you for pointing that! Fixed it a few seconds ago.

nkanazawa1989:
Thanks @0sophy1 for update! This looks good to me.

I have one more request and then let's merge this PR. Can you also update CR pulse comparison plot? Plotting CR pulse in different channels seems bit weird because _in principle_ pulse channels are not interchangable. New drawer can take matplot axis object, so you can prepare two axes to plot each waveform. In this way you can play both pulses on U1 (original pulse channel) and compare them.

This is an example code.
```python
fig, axes = plt.subplots(nrows=2, sharex=True, figsize=(8, 7))

with pulse.build(backend, name="Default CR") as cr_sched_default:
        pulse.play(cr_pulse, ControlChannel(1)) 
cr_sched_default.draw(backend = backend, axis=axes[0])
axes[0].set_xlabel("")

with pulse.build(backend, name="Custom CR") as cr_sched_custom:
    pulse.play(my_cr_pulse, ControlChannel(1))
cr_sched_custom.draw(backend = backend, axis=axes[1])
```

nkanazawa1989 (approved):
Awesome! Thanks for updating this chapter. This is the first practical example of pulse gate. @frankharkins Can you merge this PR?

frankharkins (approved):
Amazing, thank you!

PR #1139, opened 2021-06-03
1.3 Fixed explanation of running circuits using statevector simulation

pulkitvyas08:
# Changes made

Changed "...we use `execute` to run our circuit..." to "...we use `run` to execute our circuit..."

# Justification

The run() method of statevector simulation is used to execute circuit but the opposite was written in the explanation of the code


frankharkins (approved):
Thank you!

PR #1140, opened 2021-06-04
6.5 Changed to Builder syntax with Parameter class

kUmezawa:
# Changes made

Changed how to write pulse schedule to Builder syntax with Parameter class

# Justification


nkanazawa1989:
Thanks @kUmezawa . This is great! Using parameters allows us to write experiments more efficiently :)

Please consider following updates

1. cell 5
This only defines `Play(GaussianSquare...)`. I think you can directly integrate this block into cell7 without using call instruction.

2. cell 11
This only modifies pulse amplitudes. You can parametrize amplitude as well and assign both frequency and amplitude before submitting the schedule. You don's need to define the same schedule twice.

3. cell 15
```python
        pulse.set_frequency(f0, meas_chan)
```
I guess you set wrong f0. The horizontal axis is shown with offset of 6.993e6.
image

This is why you have frequency shift of 6.993 GHz in schedule plot. We cannot drive measure channel at kHz frequency.

I don't think
```python
        pulse.set_frequency(backend_defaults.qubit_freq_est[qubit], drive_chan)
```
is necessary because drive frequency is automatically adjusted to `backend_defaults.qubit_freq_est[qubit]` when you don't set frequency.

```python
        #meanwhile do nothing with the resonator
        pulse.delay(get_closest_multiple_of_16(pulse.seconds_to_samples(drive_duration)), meas_chan)
        #wait for the duration of the time delay before measuring the qubit
        pulse.delay(get_closest_multiple_of_16(pulse.seconds_to_samples(delay)), meas_chan)
```
You can use [delay_qubits](https://github.com/Qiskit/qiskit-terra/blob/e122c9c0cef78d1ba4ac57442cd03fb0363ba93c/qiskit/pulse/builder.py#L2082-L2084) function here.

```python
        #sequence for measuring the qubit
        with pulse.build(backend=backend) as temp_sched:
```
Instead of nesting the schedule, you can use `with pulse.align_left():` here to left align measure and acquisition.


Can you add warning cell to say we cannot run this experiment on this device now? And please print backend version in the cell (IIRC you can get this with `backend.version`). Once we update the controller configuration we the backend version will be incremented. We can remove this warning after the hardware issue is fixed.

kUmezawa:
@nkanazawa1989 
Thank you for your review. I will fix it.

nkanazawa1989:
Can you remove this line too?
```
backend.configuration().parametric_pulses = []  # Will allow us to send a larger waveform for our experiments
```

This is bit weird. We confirmed removing parametric pulse doesn't help to run larger waveform. Basically this is limited by the hardware waveform memory, and both parametric pulse and waveform are converted into the same data array at backend compiler.

nkanazawa1989 (approved):
LGTM! Thanks @kUmezawa !

frankharkins (approved):
Looks good, thank you!

PR #1149, opened 2021-06-08
Changing shebang in `build` and `deploy` scripts for better portability

paniash:
# Changes made

Changed shebang in `build.sh` from `#!/bin/bash` to `#!/usr/bin/env bash`.

# Justification

In `build.sh`, the bash interpreter path is provided as `#!/bin/bash`, which is good but only for specific systems (*nix). In other systems, this path might not be where `bash` exists and a reasonable shebang would be `#!/usr/bin/env bash`, which is much more portable. Please refer [this](https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash/16365367#16365367) to learn more.


divshacker:
Have you tested this on other systems? @paniash

paniash:
Explicitly no. But see no reason why it shouldn't work. :-)

frankharkins (approved):
after investigating, I'm approving; thank you!

PR #1156, opened 2021-06-10
[1.3[2.1]] Added a new Reminder Section(Conjugate Transpose)

### LINKED ISSUE #870: [2.1] Minor improvement to section 2.1
**Describe the issue**
Should mention that |> is a column vector and define complex transpose.  The * notation is not defined. SEE CAPS BELOW
A Very Important Rule 

There is a simple rule for measurement. To find the probability of measuring a state 
|ψ⟩
 in the state 
|x⟩
 we do:
p(|x⟩)=|⟨x|ψ⟩|2
The symbols 
⟨ and | tell us ⟨x| is a row vector AND |> iS A COLUMN VECTOR.  In quantum mechanics we call the column vectors kets and the row vectors bras. Together they make up bra-ket notation. Any ket 
|a⟩
 has a corresponding bra 
⟨a|
, and we convert between them using the conjugate transpose.

Reminder: The Inner Product (Click here to expand).  CHANGE TITLE TO CONJUGATE TRANSPOSE
There are different ways to multiply vectors, here we use the inner product. The inner product is a generalisation of the dot product which you may already be familiar with. In this guide, we use the inner product between a bra (row vector) and a ket (column vector), and it follows this rule:

ADD DEFINE for *. AS IN IT CHANGES THE SIGN OF THE COMPLEX PART OF THE NUMBER

shil-m:
Closes #870
Added the definition of conjugate transpose under a new Reminder Section
# Changes made

The section marked was added.
image

# Justification

The knowledge of inter conversion between bra-ket and hence the knowledge of conjugate transpose is fundamental for understanding "The inner product" as described by the issue. 
I am open to suggestions/review regarding the content and the placement.


shil-m:
@frankharkins  Hi, I didn't run the ipynb file in my local system before pushing, not sure why the output format change to png.

frankharkins (approved):
Looks good, thank you!

PR #1157, opened 2021-06-10
Closes #1152 Replacing 'up/down' with 'along'

### LINKED ISSUE #1152: [8.1] Vectors: "Down" the x-axis
**Describe the issue**
Section: Vectors and Vector Spaces
`This vector can be visualized as an arrow pointing in the direction of  3 units down the x axis and 5 units up the y axis:`
I think using "down" the x axis is ambiguous. It doesn't really tell about a particular direction especially when we are using "up" the y axis. So, it sounds a bit confusing.

In my suggestion I think it is better to use something more general like "along" the x axis and "along" the y axis.

I am open to suggestions. I am looking forward to make the changes.
Screenshot (69)_LI

shil-m:
[8.1 Vectors and Vector Spaces]: Replaced up and down with along
# Changes made

Before: `This vector can be visualized as an arrow pointing in the direction of 3 units down the x axis and 5 units up the y axis:`
After: `This vector can be visualized as an arrow pointing in the direction of 3 units along the x axis and 5 units along the y axis:`

# Justification

As described in issue #1152 using "up" and "down" creates ambiguity. x-axis and y-axis by default refer to positive direction. So this modification gives more clarity.


frankharkins (approved):
Looks good thanks!

PR #1162, opened 2021-06-13
4.1.3, 5.4 Fixed the code by reversing the string x

riya-m:
Resolved the issue by reversing the vector.

# Changes made

# Justification


frankharkins:
@riya-m can you explain how this fixes the code please? And how this affects the results?

riya-m:
@frankharkins in the cost function of section 5.4 of chapter 4.1.3 the first bit of the vector 'x' refers to the last qubit that is the 4th qubit, the second bit to the last second qubit and so on.  
I reversed the vector 'x' to fix this issue and it does not affect the plot in this particular example as the graph is symmetric.

frankharkins (approved):
Thanks!

frankharkins (approved):

PR #1167, opened 2021-06-17
1.4 Updated the Single Qubit Gates section according to the latest update

### LINKED ISSUE #1163: "Single Qubit Gates" section seems outdated
**Is your feature request related to a problem? Please describe.**

The last two paragraphs in "Single Qubit Gates" section are:

> Before running on real IBM quantum hardware, all single-qubit operations are compiled down to  U1,  U2  and  U3. For this reason they are sometimes called the physical gates.

> It should be obvious from this that there are an infinite number of possible gates, and that this also includes Rx and Ry-gates, although they are not mentioned here. It must also be noted that there is nothing special about the Z-basis, except that it has been selected as the standard computational basis. That is why we have names for the S and T-gates, but not their X and Y equivalents (e.g.  √X [...]

My notes:
- U1, U2, and U3 are deprecated.
- This [page](https://quantum-computing.ibm.com/services), displays a different set of gates as the physical gates.
- Qiskit provides the square root of X a name since version 0.20.0 (SX gate)

**Describe the solution you'd like**
The two paragraphs need to be up to date.

tgag17:
# Changes made

Changed the U3 gate to the U gate and replaced the explanation of the U1 and U2 gates with examples related to them (deriving it from the U gate).
Added a small description about the SX and SXdg gate and changed the list of physical gates used in real hardware runs.
Fixes #1163 

# Justification

The latest update does not have the U1 and U2 gates and the U3 has been replaced with the U gate.
The list of physical gates has also been updated along with the addition of the SX and SXdg gates


frankharkins (approved):
Looks good, thanks! FYI, I have also updated the widget code, this was pretty much just a search / replace for `rz` -> `p` and `Rz` -> `P` (and a slight cleaning up of code while I was there).

PR #1181, opened 2021-06-25
Removing Content Bugs caused by replacing Qasm simulator with Aer Simulator

### LINKED ISSUE #1180: Update text referring to qiskit simulators
**Describe the issue**

Since the introduction of `AerSimulator` as the preferred simulator, we have some text in the textbook that doesn't match with the code. For example:

> This behaviour is not ideal for learning, so Qiskit provides different quantum simulators: The `qasm_simulator` behaves as if you are interacting with a real quantum computer, and will not allow you to use `.get_statevector()`. Alternatively, `aer_simulator`, (which we have been using in this chapter) does allow peeking at the quantum states before measurement, as we have seen.

Here I think '`aer_simulator`' has been added through find-all-replacing statevector simulator, but the paragraph still feels slightly off and I think we're meant to use `aer_simulator` over `qasm_simulator`.

This text (and any others that refer to the simulators) should be updated. This page is a good reference: https://qiskit.org/documentation/tutorials/simulators/1_aer_provider.html#The-Aer-Simulator

divshacker:
# Changes made

fixes #1180 

# Justification

All details of this PR can be found in above linked issue

### DIFF
--- a/content/ch-appendix/qiskit.ipynb
+++ b/content/ch-appendix/qiskit.ipynb
@@ -764,7 +764,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Classical registers and the qasm simulator"
+    "### Classical registers and the Aer simulator"
    ]
   },
   {


content/ch-applications/vqe-molecules.ipynb
```diff
@@ -206,7 +206,7 @@
    "outputs": [],
    "source": [
     "from qiskit import Aer, transpile, assemble\n",
-    "backend = Aer.get_backend(\"qasm_simulator\")\n",
+    "backend = Aer.get_backend(\"aer_simulator\")\n",
```
frankharkins, line 209: This notebook (`vqe-molecules.ipynb`) isn't updated to the latest version of Qiskit, I think we should revert the changes in this chapter so the code can still be run using Qiskit 0.23.0. Otherwise, the chapter doesn't work for either.

frankharkins (changes requested):
Just one comment, otherwise looks good- thanks!

content/ch-applications/vqe-molecules.ipynb
divshacker: Ok let me change it

divshacker:
Done @frankharkins

frankharkins (approved):
Thank you!

PR #1182, opened 2021-06-25
Representing -qubit -states.ipynb

rajat709:
Representing qubit states(1.3);
in para (1.3) exploring qubits with qiskit last code has showing error:
code line = 2173

given:
```
qobj = assemble(qc)
results = sim.run(qobj).result().get_counts()
plot_histogram(results)
```

correct:
```
qobj = assemble(qc)
results = sim.run(qobj).result().get_counts()
plot_histogram(result)
```

 { 's' is extra in 'result' }

# Changes made

# Justification

### DIFF
--- a/content/ch-states/representing-qubit-states.ipynb
+++ b/content/ch-states/representing-qubit-states.ipynb
@@ -2170,7 +2170,7 @@
    "source": [
     "qobj = assemble(qc)\n",
     "results = sim.run(qobj).result().get_counts()\n",
-    "plot_histogram(results)"
+    "plot_histogram(result)"
    ]
   },
   {


frankharkins:
Thanks for making a PR. From what I can tell, in these two original lines, one assigns the variable '`results`' and the other plots '`results`'.
```
results = sim.run(qobj).result().get_counts()  # assign the variable 'results'
plot_histogram(results)  # plots the histogram of 'results'
```
I think this must be correct because it's a pretty common pattern to get results and plot them immediately. When you run the updated code, do you still get sensible outputs or do you get an error?

rajat709:
Ya, ok update has rectified the bug.

frankharkins:
Ok I will close this, thanks

PR #1190, opened 2021-06-26
2.3 Update sign in superdense coding

### LINKED ISSUE #1064: inconsistency superdense coding chapter
in the last line of step 3 - bob's operations:
-|11>+|10>   gives -|11> after H operation on rightmost qubit.

I think it should be 
+|11>

also in the explanation text, for message "11", operation ZX is applied. in the python code XZ is applied though.

clausia:
# Changes made

In step 3 of the superdense coding protocol, when Bob decodes the message, at the end he must apply an H gate to Alice's qubt, when doing this for the fourth message, the final state is | 11>, that is, it is positive and not negative as it currently is.

I have modified the image of the table, as well as eliminated the paragraph that tries to justify the negative sign, since it is not necessary anymore.

# Justification

I want to resolve #1064

I include in the following images the demonstration that the negative sign is incorrect, and for a demonstration with matrix multiplication please see [here](https://github.com/clausia/proof-fix-superdense-coding-qiskit-textbook/blob/main/Superdense%20Coding%20fix.ipynb).

image
image


clausia:
I realized myself about this little mistake and decided to fix it, I have made my own test that in fact the last case should not have a minus sign. At the time of creating the commit and the PR I looked for if there was an issue in this regard, when I found that there was, I added the reference to it (there were no more comments since May)

content/ch-algorithms/superdense-coding.ipynb
```diff
@@ -118,12 +118,12 @@
...
     "| $|01\\rangle + |10\\rangle$  | $|11\\rangle + |10\\rangle$  |  $|10\\rangle$   |\n",
     "| $|00\\rangle - |11\\rangle$  | $|00\\rangle - |01\\rangle$  |  $|01\\rangle$   |\n",
-    "| $-|01\\rangle + |10\\rangle$ | $-|11\\rangle + |10\\rangle$ | $-|11\\rangle$   |\n",
+    "| $-|01\\rangle + |10\\rangle$ | $-|11\\rangle + |10\\rangle$ |  $|11\\rangle$   |\n",
     "-->\n",
     "\n",
     "![image3](images/superdense_table2.png)<br>\n",
     "\n",
-    "When Bob performs the measurements, he is certain to get the message Alice sent him. Note that the last outcome has a negative sign $-|11\\rangle$, writing it as $|11\\rangle$ is also correct as this is global phase and is not measurable.\n",
+    "\n",
```
divshacker, line 126: I think this is unnecessary extra space.

frankharkins (approved):
Looks good, thank you @clausia and @divshacker

PR #1204, opened 2021-07-08
 4.1.3 - Improve QAOA Chapter

amitracal:
Co-authored by Abhijit Mitra and Vaibhaw Kumar

# Changes made

QAOA Tutorial Rewritten with simpler examples and explanation, we explained a max-cut problem through new images and then added runnable code to solve the max-cut using QAOA, the older version of the tutorial is kept as advanced material as appendix.

# Justification

The previous chapter was less accessible and needed more advanced understanding of the topic and the code was outdated.


frankharkins (approved):
Awesome, thank you both!

PR #1206, opened 2021-07-13
3.10 Get_subsystems_counts for histogram of teleported qubit

splch:
# Changes made

I used the get_subsystems_counts method to make seeing the teleported qubit's state more clear.

# Justification

Currently, the teleported state is not clearly seen as all three measured qubits appear in the histogram as bitstrings. This proposed change uses the qiskit.aqua.utils' module get_subsystems_counts to display individual qubit counts. Now, readers can identify that the blue bar is always |0> without having to look across the bitstrings and confirm all q_2 measurements are 0.

Before:
image
After:
image

If you'd like only the qubit_2 to display in the histogram, I can change that by:
```
t_qc = transpile(qc, sim)
t_qc.save_statevector()
counts = get_subsystems_counts(sim.run(t_qc).result().get_counts())[0]
plot_histogram(counts)
```


frankharkins:
Sorry it took a while to review this. I like this idea but I see you're using Aqua, which is being deprecated. I can't seem to find `get_subsystems_counts` in any other part of Qiskit, so I suggest we use [`marginal_counts`](https://qiskit.org/documentation/stubs/qiskit.result.marginal_counts.html) instead.

frankharkins (changes requested):
Sorry it took a while to review this. I like this idea but I see you're using Aqua, which is being deprecated. I can't seem to find `get_subsystems_counts` in any other part of Qiskit, so I suggest we use [`marginal_counts`](https://qiskit.org/documentation/stubs/qiskit.result.marginal_counts.html) instead.

splch:
> Sorry it took a while to review this. I like this idea but I see you're using Aqua, which is being deprecated. I can't seem to find `get_subsystems_counts` in any other part of Qiskit, so I suggest we use [`marginal_counts`](https://qiskit.org/documentation/stubs/qiskit.result.marginal_counts.html) instead.

Sounds great! I'll use `marginal_counts` instead. I can iterate through `marginal_counts` to recreate `get_subsystems_counts`, but I think I'll submit a pull request that returns a default behavior closer to that of `get_subsystems_counts`.

This is what I plan on doing:
```python
t_qc = transpile(qc, sim)
t_qc.save_statevector()
counts = sim.run(t_qc).result().get_counts()
qubit_counts = [marginal_counts(counts, [qubit]) for qubit in range(3)]
plot_histogram(qubit_counts)
```
image

frankharkins (approved):
Thank you!

PR #1207, opened 2021-07-13
1.4 Add U-gate example with statevector

splch:
# Changes made

This change adds a basic U-gate example performing the same operation as an H-gate. In addition, the statevector simulator shows the qubit in the identical state as the H-gate. In this way, readers can more easily identify the U-gate as a universal gate.

# Justification

Every single-qubit gate had a circuit example and visualization showing it's function except the U-gate.  The page is now more consistent as well as slightly more interesting since readers will hopefully try to create other gates with the U-gate.

Before:
image
After:
image


frankharkins (approved):
Thank you!

PR #1212, opened 2021-07-15
Typo, Latex + List display error on QAOA page

StreakSharn:
## 4.1.3 Fixed misspelling of 'Hamiltonian', error in Latex formula and error in bullet point display.

# Changes made

In https://qiskit.org/textbook/ch-applications/qaoa.html, there is a typo - "To construct such a Hamitlonian" instead of Hamiltonian, just above 'The Mixing Unitary' heading. Further down, under 'The QAOA Circuit', there is a Latex error where the formula is incorrectly displayed: 

image

 
Under the 'Problems' heading, the bullet point list is also inconsistently displayed, as one part ('We have given a corrext expression...') used dashes whereas the other correctly used the bullet points.

image


frankharkins (approved):
Awesome, thanks!

PR #1219, opened 2021-07-19
Interactive Bloch Sphere

alihakimtaskiran:
I made a interactive bloch sphere. You are able to rotate the bloch sphere now

### DIFF
--- a/content/ch-states/representing-qubit-states.ipynb
+++ b/content/ch-states/representing-qubit-states.ipynb
@@ -150,6 +150,7 @@
    },
    "outputs": [],
    "source": [
+    %matplotlib notebook
     "from qiskit import QuantumCircuit, assemble, Aer\n",
     "from qiskit.visualization import plot_histogram, plot_bloch_vector\n",
     "from math import sqrt, pi"


frankharkins:
@alihakimtaskiran is there a mistake? In 'files changed' you have only added `%matplotlib notebook`, I'm not sure why that's necessary and I can't see anything relating to an interactive Bloch sphere...

alihakimtaskiran:
@frankharkins As we attempt to print bloch sphere we only see picture of the bloch sphere. But `%matplotlib notebooks` enables JavaScript backend. The backend makes the bloch sphere interactible. We can rotate it.

frankharkins:
Ok, just tried it. I'm not sure we should add this because:
1. It affects all images, not just the Bloch spheres. I think the extra controls and information is unnecessary for a circuit diagram. 
Screenshot 2021-07-28 at 12 45 26

2. I see with `%matplotlib inline` we can now rotate the sphere, but I'm not sure it adds much educational value. I don't think being able to rotate the sphere helps with understanding.

Open to other opinions on this.

content/ch-states/representing-qubit-states.ipynb
```diff
@@ -150,6 +150,7 @@
    },
    "outputs": [],
    "source": [
+    %matplotlib notebook
```
frankharkins, line 153:
```suggestion
   "%matplotlib notebook\n",
```
If we are to add this, it needs to be valid JSON to be part of the notebook.

frankharkins:
I'm going to close this for the reasons stated above. Feel free to reopen if you disagree.

PR #1224, opened 2021-07-22
[3.5] Standardized ket notation in QFT section

### LINKED ISSUE #1223: Inconsistent ket notation in Quantum Fourier Transform
**Describe the issue**
In Section 6 of the Quantum Fourier Transform section, the wavefunctions in the equations don't use the ket notation that previous sections use.

The same ket notation used in previous sections is used consistently throughout the page.

mohitd:
Fixes #1223

# Changes made
Updated the wavefunction notation in the 3-qubit example to use the same ket notation as previous sections.

# Justification
The notation used in previous sections always wrote the wavefunction psi as a ket to distinguish it as a quantum state (as opposed to some a normal variable).


frankharkins (approved):
Thank you!

PR #1229, opened 2021-07-25
[1.2] Fixed a calculation mistake and added clarity to describe CNOT gate

suraniakbar:
Replaced 64 with 128 as 2^7 is 128. We are using 8 bits to represent 2^0 to 2^7.


frankharkins (approved):
Awesome, thank you!

PR #1231, opened 2021-07-25
[3.7] Typo?

WhiteBlackGoose:
3rd chapter, 3.7.

Is it correct that we actually have 8 qubits?

image


frankharkins (approved):
Whoops, thanks!

PR #1234, opened 2021-07-25
3.2 A section instructs using incorrect function

suraniakbar:
Quick exercise suggests using plot_bloch_sphere_spherical() function. The correct function to use here is plot_bloch_vector_spherical().


frankharkins (approved):
Thank you!

PR #1249, opened 2021-07-28
Update hamiltonian-tomography.ipynb(version table)

rajat709:
# Changes made

# Justification

### DIFF
--- a/content/ch-quantum-hardware/hamiltonian-tomography.ipynb
+++ b/content/ch-quantum-hardware/hamiltonian-tomography.ipynb
@@ -993,23 +993,21 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 30,
+   "execution_count": 52,
    "metadata": {},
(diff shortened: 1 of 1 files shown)


frankharkins:
@rajat709 thanks for your contribution. We have these cells to let the reader know which version of Qiskit was used to produce the outputs, so the reader can reproduce the results. Since you've only run the final cell, the version information won't match the version the rest of the notebook was run with.

This means you/we need to (a) re-run the entire notebook or (b) wait until we end up running it for some other reason. If you want to go with (a), then please use the latest version of Qiskit (I think 0.28.0, you can do `pip install --upgrade qiskit` to make sure).

frankharkins:
Closing RE [the comment above](https://github.com/qiskit-community/qiskit-textbook/pull/1249#issuecomment-891127110). Please open another PR if you still want to work on this. Thanks.

PR #1261, opened 2021-08-12
[6.4] fixes content bug

### LINKED ISSUE #799: 6.4 where does sigma_x come from?
**Describe the issue**
1. Where does sigma_x in the last equation come from? Before we were dealing just with sigma_z.

2. In equation 2 it is not clear why from the commutator of sigma_+- terms the result is sigma_z. What is the [sigma_+, sigma_-]. It would be good to mention this in the textbook.

6_4_sigma_x

divshacker:
# Changes made

fix #799 

# Justification

All details of this PR can be found in above linked issue


quantumjim:
Is it definitely correct that it is just the x -> z change that is needed? My first thought was that \sigma_x came from \sigma_+ and \sigma_-. I think @nbronn should review.

nbronn (commented):
From 2 lines above, it appears the Pauli term comes from $\sigma^- + \sigma^+ \equiv \sigma^x$, so the previous version seems correct with me. Checking against Eq. 2.72 of Lev Bishop's [thesis](https://www.levbishop.org/thesis/Bishop-thesis.pdf) seems to confirm this within a factor of 2.

divshacker:
Please let me know what else can be added

frankharkins (commented):
Looks good to me, just need @nbronn to approve latest commits

nbronn (approved):
Looks good @divshacker

frankharkins (approved):
Thanks!

PR #1262, opened 2021-08-13
[9.4 Variational Quantum Regression] Fix missing image not showing in textbook

clearya1:
# Changes made

Changed  image syntax to <src > syntax so the image can render on online textbook

# Justification

The image wasn't showing on the online textbook


frankharkins:
Thanks for taking a look! This was fixed in #1260 but hadn't been pushed to the live site until today. Sorry for the miscommunication.

PR #1267, opened 2021-08-18
[3.11] Normalized qubit ordering for Superdense Coding

### LINKED ISSUE #1266: [3.11] Inconsistent qubit ordering within Quantum Teleportation and Superdense Coding
**Describe the issue**
In the previous section on Quantum Teleportation, in the math, Alice's qubit is first in the pair while Bob's qubit is the second in the pair. Earlier, we stated that Qiskit uses reverse ordering so the code example still makes sense. However, the ordering for the math is flipped in the Superdense Coding where it uses the same ordering as the code example (Alice's qubit is the second), which also flips the canonical gates (e.g., Alice uses the X gate to send 10 instead of 01).

The qubit ordering for both Quantum Teleportation and Superdense Coding should be the same for the math so the two articles are consistent with each other.

mohitd:
Fixes #1266

# Changes made
Flipped qubit ordering so Alice's qubit is first in the pair and Bob's is second. Corrected all equations and code to match. Also added an HTML table to replace the image for easier editing.

# Justification
The section on Quantum Teleportation used the ordering that Alice's qubit should come first. Since Superdense Coding is very closely related, for consistency, it should use the same qubit ordering


frankharkins (approved):
Thanks for pointing this out. FYI, while reviewing:
- I fixed the code so it gives the right output (and we don't need to explain the reversed results after)
- I fixed some problems later down in the notebook caused by the change
- I re-ran the cells to get fresh output
I also added some docstrings and neatened up the code a bit, I've been meaning to do this for a while and decided to do it now while I was in the headspace.

PR #1271, opened 2021-08-20
Put period inside displayed formula

jtweigle:
# Changes made

I noticed a period outside a displayed formula where the formula is at the end of the sentence, so I moved it inside.

# Justification

The change is just a fix of punctuation.


frankharkins (approved):
thanks!

PR #1272, opened 2021-08-20
Make the name for controlled U gates consistent

jtweigle:
In one place, the controlled U gate is called CU. In three others, it was called C-U. Here I’m changing them to all say CU.

# Changes made

I changed three occurrences of C-U (for a controlled U gate) to CU, to be consistent with the first time the CU is introduced.

# Justification

My decision here is to change C-U to CU, because the hyphen was rendering in LaTeX math as a minus. That was a source of confusion for me as a first-time reader: it looked like some gate C, minus some gate U applied 2^j times. On the other hand, if a hyphen is preferable for some other reason I don’t know about, then `C\text{-}U` would be a good option.

### DIFF
--- a/content/ch-algorithms/quantum-phase-estimation.ipynb
+++ b/content/ch-algorithms/quantum-phase-estimation.ipynb
@@ -99,7 +99,7 @@
     "\n",
     "\n",
     "\n",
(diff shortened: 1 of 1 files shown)


frankharkins (approved):
Thank you!

PR #1274, opened 2021-08-21
[5.5] Minor math bug fix

### LINKED ISSUE #1273: conjugate transpose of Pauli Y matric is wrong
Please refer to [this section of The Density Matrix & Mixed states](https://qiskit.org/textbook/ch-quantum-hardware/density-matrix.html#3.-Properties-of-the-Density-Matrix--). in the example under "Unitary Evolution", conjugate transpose of Pauli Y matrix is wrong. 

It should be `[[0, -i],[i,0]]` in place of `[[0, i],[-i,0]]` because Pauli's matrices are Hermitian which means they are their own conjugate-transpose.
screenshot

divshacker:
# Changes made

fixes #1273 

# Justification

All details can be found in above linked issue. 
I did an extra step in fixing this bug as when the  conjugate transpose was wrong so the multiplication was also wrong as it was using that wrong matrix. I fixed that too.


mayankkaizen (dismissed):

mayankkaizen:
While you are at it, I suggest one more edit. Further down the relevant section, the same multiplication was done using Qiskit software which produced the correct result. Then there was this remark -

>As expected, this result matches our calculation of ρ′0 up to a global phase of -1.

Obviously this is also wrong because it is being compared to the earlier wrong result.

divshacker:
> While you are at it, I suggest one more edit. Further down the relevant section, the same multiplication was done using Qiskit software which produced the correct result. Then there was this remark -
> 
> > As expected, this result matches our calculation of ρ′0 up to a global phase of -1.
> 
> Obviously this is also wrong because it is being compared to the earlier wrong result.

Hey, I have removed that line as the global phase is 1 and it is not worth mentioning that line in this chapter now.

frankharkins (approved):
Awesome, thank you!

PR #1202, opened 2021-07-06
Quantum Coin Game and TSP with QPE

pritamsinha2304:
# Changes made
Added new notebook on quantum coin game and travelling salesman problem with quantum phase estimation.

PR #1282, opened 2021-09-11
Added Quantum Coin Game

pritamsinha2304:
Added quantum coin game notebook separately in response to pull request #1202 

# Changes made
1> Added references
2> Compacted the code cell
3> Removed some visualizations (e.g. Bloch sphere )

# Justification
These changes has been asked by @frankharkins.


content/quantum_coin_game/quantum_coin_game.ipynb
```diff
@@ -0,0 +1,1071 @@
...
+    "### **Analogy** <a id=\"analogy\"></a>\n",
+    "\n",
+    "---\n",
+    "\n",
+    "Now that we know what is a quantum coin game, what is it based on and most importantly what are the rules of this game, lets convert the concept of this game in quantum computing terminology.\n",
+    "\n",
+    "* The 'coin' in flipping a coin we referring here is a '**single qubit gate**'.\n",
+    "$$\n",
+    "      |\\psi\\rangle=\\begin{bmatrix}\\alpha \\\\ \\beta\\end{bmatrix}\n",
+    "$$\n",
```
divshacker, line 98: I think this equation is not rendering properly.

divshacker:
Also I think one should use `Aer_simulator ` instead of `QASM` to make it more general way. Just a suggestion 
What do you say @frankharkins ?

content/quantum_coin_game/quantum_coin_game.ipynb
pritamsinha2304: Working for me

divshacker: If it is working in your local system then leave it as it will automatically fixed when pushed to website

divshacker:
Also If you add some images showing your approach of applying Quantum superposition in classical coin game. Then it will be more clear for readers. Just a suggestion.

frankharkins (changes requested):
I still think this content can be trimmed down significantly; you explain the concept of the game very well in “approach”:

> The human plays, choosing whether to flip the coin (apply the 𝑋 operator). However, since the 𝑋 operator just flips the state vector upside down, 𝑋 has no effect.

Since we already know the best strategy, the next section (and a lot of the code example) is redundant. The code is also a bit convoluted, e.g. we go through a dict (`quantum_moves`) and two functions (`random_moves` & `corresponding_gate`) to choose which gate to apply.

I would suggest linking to Sohini’s talk in the introduction and reframing the page as an explanation to accompany her talk. I think it would be better to stick to the optimal quantum computer strategy and focus on reproducing her claims. You could allow the reader to “play” and input their decision using `input` or a [widget](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Button).

It would also be nice to include a device run to show where the “97%” comes from, although I think today’s quantum computers would do better than that.

pritamsinha2304:
> I still think this content can be trimmed down significantly; you explain the concept of the game very well in “approach”:
> 
> > The human plays, choosing whether to flip the coin (apply the 𝑋 operator). However, since the 𝑋 operator just flips the state vector upside down, 𝑋 has no effect.
> 
> Since we already know the best strategy, the next section (and a lot of the code example) is redundant. The code is also a bit convoluted, e.g. we go through a dict (`quantum_moves`) and two functions (`random_moves` & `corresponding_gate`) to choose which gate to apply.
> 
> I would suggest linking to Sohini’s talk in the introduction and reframing the page as an explanation to accompany her talk. I think it would be better to stick to the optimal quantum computer strategy and focus on reproducing her claims. You could allow the reader to “play” and input their decision using `input` or a [widget](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Button).
> 
> It would also be nice to include a device run to show where the “97%” comes from, although I think today’s quantum computers would do better than that.

So, if I undetstood it correctly I should only include that part where the quantum computer win 97% of the time or quantum computer with mind part.
If so should I delete the previous section ? 
Confirm please.

pritamsinha2304:
@frankharkins I have updated the notebook.

frankharkins:
I think this is great, really like the direction it's going in. Things that need changing are:
- `res_count` is undefined, I think you just need to change this to `res`
- There are quite a few unused imports
- You don't need to include the `pip install...` cell as it should be installed already on the website and in quantum lab

Some other ideas (these are subjective so let me know what you think):
- I like the widget you've included, would it make sense to have the whole example in one or two code blocks? It might feel more like you're "playing" against the computer if you can select a gate then press "run" once.
- I suggest picking a different device, Armonk is quite old now and I got much better results with `ibmq_santiago`. Reason for this is to show more recent device performance.

pritamsinha2304:
@frankharkins I have updated the notebook with those suggestions. I have added a separate "play it" content. See which one to keep.

content/quantum_coin_game/quantum_coin_game.ipynb
```diff
@@ -0,0 +1,3911 @@
...
+    "\n",
+    "\n",
+    "### What is Quantum Coin Game ? <a id=\"quantum_definition\"></a>\n",
+    " Quantum Coin Game is one of the fundamental concept of quantum computing, which uses simple implementation of quantum gates or more precisely uses the wierdness of quantum mechanics, to win about 97% of the time, when played against an opponent. Flipping of coin and say heads or tails.\n",
+    "\n",
+    "### Where the concept came from ? <a id=\"concept\"></a>\n",
+    " The concept of Quantum Coin Game came from the idea of classical coin game which can only show heads and tails. But since the game utilizes the concepts of quantum mechanics, it would be interesting to see what could be the outcome of the whole experiment.\n",
+    "\n",
+    "### What is the main idea of this game ? <a id=\"idea\"></a>\n",
+    " The main concept of this game is how the quantum computer uses the power of **QUANTUM SUPERPOSITION**, which tells an object can exists in 2 different states at the same time, to win absolutely everytime. The fact is an quantum computer can do superposition already puts it in a higher chance of winning over the classical flip.\n",
```
frankharkins, line 60:
```suggestion
    " The main concept of this game is how the quantum computer uses the power of quantum superposition, which tells an object can exists in 2 different states at the same time, to win absolutely every time.\n",
```

content/quantum_coin_game/quantum_coin_game.ipynb
```diff
@@ -0,0 +1,3911 @@
...
+    "\n",
+    " **NOTE**: To learn more about quantum superposition, link to \"[Qiskit Textbook](https://qiskit.org/textbook/ch-states/representing-qubit-states.html)\" superposition page.\n",
+    "\n",
+    "### What are the rules of this game ? <a id=\"rules\"></a>\n",
+    " 1. Quantum Computer plays a move but it is not revealed to the Opponent(Human).\n",
+    " 2. Opponent(Human) plays a move and it is also not revealed to the Quantum Computer.\n",
+    " 3. Finally Quantum Computer plays a move.\n",
+    " 4. Results are shown. If its heads, then Quantum Computer wins. Else, Opponent(Human) wins.\n",
+    " \n",
+    "**NOTE**: \"Playing a move\" refers to \"Flipping the coin\" and we consider the coin as **FAIR COIN**.\n",
```
frankharkins, line 70:
```suggestion
    "**NOTE**: \"Playing a move\" refers to \"Flipping the coin\" and we consider the coin as fair coin.\n",
```

content/quantum_coin_game/quantum_coin_game.ipynb
```diff
@@ -0,0 +1,3911 @@
...
+    "$$\n",
+    "  \\frac{0}{2} = 0 \\%\n",
+    "$$"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This table shows the quantum computer wins $100\\%$ of the time. But actually it is $~97\\%$, due to certain error."
```
frankharkins, line 253:
```suggestion
    "This table shows the quantum computer wins $100\\%$ of the time. But in Shohini's talk it is $~97\\%$, due to errors."
```

content/quantum_coin_game/quantum_coin_game.ipynb
```diff
@@ -0,0 +1,3911 @@
...
+    "\n",
+    "### What are the rules of this game ? <a id=\"rules\"></a>\n",
+    " 1. Quantum Computer plays a move but it is not revealed to the Opponent(Human).\n",
+    " 2. Opponent(Human) plays a move and it is also not revealed to the Quantum Computer.\n",
+    " 3. Finally Quantum Computer plays a move.\n",
+    " 4. Results are shown. If its heads, then Quantum Computer wins. Else, Opponent(Human) wins.\n",
+    " \n",
+    "**NOTE**: \"Playing a move\" refers to \"Flipping the coin\" and we consider the coin as **FAIR COIN**.\n",
+    "\n",
+    "**NOTE**: Refer to [Sohini's Ted Talk](#conclusion)\n",
```
frankharkins, line 72:
```suggestion
    "**NOTE**: Refer to [Shohini's Ted Talk](#conclusion)\n",
```

content/quantum_coin_game/quantum_coin_game.ipynb
```diff
@@ -0,0 +1,3911 @@
...
+    "res_real = job_real.result().get_counts()\n",
+    "print(res_real)\n",
+    "plot_histogram(res_real)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### **Quantum Computer Wins all the time.**"
```
frankharkins, line 553:
```suggestion
    "Unlike the perfect simulation, the real quantum computer only wins ~$99\\%$ of the time, the $1\\%$ in which it loses is due to errors. Quantum computers have improved a bit since [Shohini's talk](#conclusion) where the error is closer to $3\\%$."
```

frankharkins (changes requested):
@pritamsinha2304 looks good! I've left some suggestions, please check them out and let me know what you think. Otherwise:
- I really like the "play it" cell! I think this would be a cool introduction near the top of the page, but up to you.
- @divshacker is right, please use the Aer simulator (should just be a simple drop-in replacement) as I think qasm simulator will be deprecated at some point in the future.
- Finally, can you move this file to `content/ch-demos/coin-game.ipynb` please?

pritamsinha2304:
@frankharkins I have updated the notebook as per your suggestions.

frankharkins (approved):
Thank you!

PR #1283, opened 2021-09-11
Added Travelling Salesman problem

pritamsinha2304:
Added tsp notebook separately in response to pull request #1202

# Changes made
1> Added proper references
2> Removed unnecessary code
3> Compacted the code cells
4> made code more functional
5> Used Qiskit's built-in library instead of separately writing it.

# Justification
These changes has been asked by @frankharkins.


divshacker:
The Circuit is too large. Can it be small ?
Also values of phi are not visible Can you fix it ?
Screenshot from 2021-09-14 09-15-48

pritamsinha2304:
> The Circuit is too large. Can it be small ?
> Also values of phi are not visible Can you fix it ?
> Screenshot from 2021-09-14 09-15-48

No.
It is working for me.

frankharkins (changes requested):
Thank you! Looking good. Some more suggestions:
- Quite a lot of time is spent explaining the outer product and adding matrices, can we link to the linear algebra page in the appendix instead?
- I think you can still reduce the number of variables a bit too, e.g. here:
   ```
   def controlled_unitary(qc, unit, eigen, x, y, z, a, b, c, d):
   ```
   could be:
   ```
   def controlled_unitary(qc, unit, eigen, qubits, phases):
   ```
   and you can then replace:
   ```
   controlled_unitary(qc, unit, eigen, m_list[0], m_list[1], m_list[2], 0, m_list[9], m_list[10], m_list[11])
   ```
   with
   ```
   controlled_unitary(qc, unit, eigen, qubits[0:3], [0]+phases[9:12])
   ```
   I also think this is more readable as the variables are named after what they represent instead of a letter / number.
- You should put "Image source [3]" and "Image source [4]" next to the appropriate images too.
- 
Some small formatting tips:
- When writing text in latex equations, use `\text{}` (e.g. `$O(n^22^n) \text{where $n$ is the number of vertex/nodes of the graph}$)`
- Bold adds emphasis, but I think there is too much, and we rarely bold text in the rest of the textbook. To keep with the style I suggest removing this. Jupyter notebook has an option to find & replace in all cells, you can replace `**` with nothing.
- Similarly for quotes (`>` & `>>`), these should be used sparingly and for actual quotes. Find and replace is a little trickier here due to HTML tags.

frankharkins:
> The Circuit is too large. Can it be small ?
> Also values of phi are not visible Can you fix it ?
> Screenshot from 2021-09-14 09-15-48

Also working fine for me, I think it will render fine on the site.

pritamsinha2304:
@frankharkins I have updated the notebook.

frankharkins (changes requested):
Hi, thanks again and sorry for taking a while to re-review. There are still quite a few changes I’d like to see. They fall under roughly two categories, the first is simplifying / shortening:
- I think the "backstory" can be removed, the user should hopefully already be familiar with this.
- The code still has lots of cases like this one:
  ```python3
  a = np.pi / 2
  b = np.pi / 8
  c = np.pi / 4
  d = np.pi / 2
  e = np.pi / 4
  f = np.pi / 4
  g = np.pi / 8
  h = np.pi / 4
  i = np.pi / 8
  j = np.pi / 4
  k = np.pi / 4
  l = np.pi / 8
  ```
 - There are a few large workings-out which puts a lot of information on the page that the reader might not need. Some examples are: 
   - The table after straight after "With 4 city, all the possible Hamiltonian cycle are-"
   - The place where you calculate |𝑎𝑎⟩⟨𝑎𝑎| for all four basis states
   - The matrices after "In similar way, we have to find 𝑈2, 𝑈3, 𝑈4"
- The circuit drawing in code cell 74 is really big. It would be more helpful if the circuit looked more like figure2 in the paper you're basing this on. You can probably do this by putting parts of the circuit into gates.
- I don't think we need to show what the QFT looks like, the reader should already know this by now.
- More generally, we want to remove any unnecessary information and keep things as concise as possible. This notebook is solving one instance of the TSP, so I think it should be treated as a short demo. You can use the `<details>` tags to hide extra workings if you think it's still needed by some readers.

The other things are to do with style:
- There are quite a few grammar errors throughout the page. This is the biggest style problem. If English is not your first language, I can help fix these.
- You should use `\text{ }` when writing words in latex, otherwise the words will be in italics and have weird kerning.
- I think there are too many level-3 headings, can you combine some sections maybe?

pritamsinha2304:
@frankharkins I have updated the notebook as per your suggestions. 
1. Same type of operation/calculation has been put within `<details>` tag.
2. Removed the backstory part and appendix part.
3. Unnecessary code cell has been removed.
4. Code has been shortened.
5. Proper representation of tables.
6. Circuit was too big, so the circuit has been compressed via `to_gate()` function and resembles that in paper.
7. Removed QFT part and added the qiskit textbook QFT link for more details.
8. Some of the level-3 headings as been removed.
9. Used `\text{}` for the word representation.
10. Fixed the table of content and references accordingly to the previous changes.


These are the changes. Let me know if this changes are correct.

frankharkins:
@pritamsinha2304 thank you! I've made a PR [here](https://github.com/pritamsinha2304/qiskit-textbook/pull/1) with some suggested changes, please review and let me know if you disagree with anything or have any questions.

pritamsinha2304:
@frankharkins Thanks for the changes. I checked those changes and it looks good. Although it looks fine here, something might break in case of formatting when it will get uploaded to qiskit textbook. When that happens, I will look at that and PR a patch for it.

frankharkins:
Ok great, if you're happy with the changes please merge the PR

PR #1287, opened 2021-09-18
[4.2.1] Fixed minor maths typos: Added missing factor of 1/2, and changed the…

quantumcharleshill:
… order of Ut and U for clarity

# Changes made

In section 3, I added a missing factor of 1/2 in the equations describing the Hadamard test where it was previously missing.
Also, I changed the order of U U\dagger to reflect the order in the previous equation. Although this is technically not incorrect, this order is clearer.

# Justification

The mathematics is incorrect without this change, as a factor of 1/2 is simply dropped from the previous line of working.


frankharkins (approved):
Thank you!

PR #1292, opened 2021-09-27
Added binary addition for number 3(in decimal)

Vaishnavi-Markunde:
Added binary addition for number 3(in decimal)  to avoid confusion of carry and sum with three '1's

# Changes made

# Justification


frankharkins:
Not sure about this, from [wikipedia](https://en.wikipedia.org/wiki/Adder_(electronics)#Half_adder), the definition of a half adder only seems to handle two input bits

PR #1299, opened 2021-10-04
update gems

vabarbosa:
# Changes made

- updated required gems to latest version
- removed `github-pages` gem 

# Justification

- there are a number of [security alerts](https://github.com/qiskit-community/qiskit-textbook/security/dependabot) which have been resolved in updated gems

- `github-pages` gem does not work with Jekyll 4.0 which is required for updates/resolution to some existing [security alerts](https://github.com/qiskit-community/qiskit-textbook/security/dependabot). deployment is done via [GitHub Action](https://github.com/qiskit-community/qiskit-textbook/blob/main/.github/workflows/deploy.workflow.yml) so the `github-pages` gem is not required


frankharkins (approved):
Amazing, thank you!

PR #1313, opened 2021-10-20
Update deutsch-jozsa.ipynb

manzanillo:
This change proposes to delete an unused var:
The shots variable is not used in the snippets. Thus I'd suggest removing it. 
Alternatively it can be added as an argument to the `backend.run()`call. But as 1024 is the default value anyway, it just makes the code more lengthy as the focus is not on the sampling number here.

# Changes made

# Justification


frankharkins (approved):
Makes sense, thank you!

PR #1321, opened 2021-10-30
[2.6] Fixed Grammar Error "to copies"

BaderTim:
# Changes made
"The effect of this is to ~~copies~~ **copy** the information over..."

# Justification
The verb after “to” should be in the base form.


frankharkins (approved):
Thank you!

PR #1323, opened 2021-11-06
[1.4, 8.1] Issue 719: enhanced explanation for measuring in different basis

### LINKED ISSUE #719: (Ch 1.4) Measuring in Different Bases: Explanation could be improved for first time readers
**Is your feature request related to a problem? Please describe.**
https://qiskit.org/textbook/ch-states/single-qubit-gates.html
Hi, I was reading the section _1.4 Single Qubit Gates_, and I think that the explanations for _Digression: Measuring in Different Bases_ and _Digression: The X, Y & Z-Bases_ quickly introduce some concepts related to bases which might not be immediately apparent to the first time reader.

In _The X, Y & Z-Bases_, the notebook states (and I quote): "In fact, the computational basis (the basis formed by the states  |0⟩  and  |1⟩ ) is often called the Z-basis. This is not the only basis we can use, a popular basis is the X-basis, formed by the eigenstates of the X-gate. We call these two vectors  |+⟩  and  |−⟩"

From the above introductory passage, it is not clear why the reader should start treating the eigenvectors of 'certain' matrices (here the X,Y and Z) to be a basis for the underlying Hilbert space. (I imagine a beginner to be thinking: Is this a claim? Or is it a general fact which can be proved? Is it only valid for particular matrices, or does this hold for all matrices in general? (of course, it doesn't hold for all matrices.)) Based on the subject matter introduced so far in the textbook, this is not very clear.

Also, in _Digression: Measuring in Different Bases_, it is stated: "As an example, let’s try measuring in the X-basis. We can calculate the probability of measuring either  |+⟩  or  |−⟩ :

p(|+⟩)=|⟨+|q⟩|^2 , p(|−⟩)=|⟨−|q⟩|^2
 
And after measurement, we are guaranteed to have a qubit in one of these two states. Since Qiskit only allows measuring in the Z-basis, we must create our own using Hadamard gates"

The code example then proceeds to demonstrate the X-basis measurement by applying a Hadamard gate before and after measurement. However, it is not obvious (from the point of view of the uninitiated) why the Hadamard is the required transformation to switch from the computational basis to the X-basis.

**Describe the solution you'd like**
Here's how I think it could be improved:
A. By explaining that it is a general property that the eigenvectors of Hermitian matrices are linearly independent, are orthogonal (for unique eigenvalues) and thus the eigenvectors of any Hermitian matrix can be used as an orthogonal basis (or at least derive one) for the underlying Hilbert space. This could initially be asserted as a claim for introducing the concept. A proof can also be added.   

B. By explaining (or at least motivating) that the Hadamard is the matrix used to translate between the eigenvectors of Z and X, and therefore it is the required matrix for measurement in the X basis.

**Describe alternatives you've considered**
The interested reader can be pointed to other resources on linear algebra, such as a textbook/webpage. 

**Additional Context**
By keeping in mind that the Qiskit textbook could be used by a broad audience, such improvements could be important to make the textbook a lucid and approachable resource.

dhruvbhq:
# Changes made

To resolve #719.

- In appendix on linear algebra (8.1), clearly stated that the eigenvectors of Hermitian and unitary operators form a basis for the underlying vector space.
- In chapter 1.4, single qubit states, added a note that since X and Y are Hermitian/unitary, their eigenvectors indeed form a basis due to the above property. (Referred it to the appendix)
- In chapter 1.4, single qubit states, added a short paragraph to show that state expressed in X-basis is obtained from state expressed in Z-basis by multiplying with Hadamard matrix.

# Justification

In the issue #719, it was discussed that the explanation of measuring in different bases could be improved and some gaps in explanation could be filled. First, it required clarifying the above property of eigenvectors of Hermitian and unitary operators. In this PR, this has been handled by adding it in the appendix dedicated to linear algebra. For enhancing the explanation for 1.4, the claim that eigenvectors of X and Y are a basis for the vector space needed to be supported by this property. A short note has been added for that. Further, in 1.4, a paragraph has been added to motivate why the Hadamard matrix translates from Z-basis measurement to one in the X-basis (as it was proposed in the issue).


content/ch-states/single-qubit-gates.ipynb
```diff
@@ -319,7 +319,7 @@
     "\n",
     "$$ |\\circlearrowleft\\rangle, \\quad |\\circlearrowright\\rangle$$\n",
     "\n",
-    "We leave it as an exercise to calculate these. There are in fact an infinite number of bases; to form one, we simply need two orthogonal vectors.\n",
+    "We leave it as an exercise to calculate these. There are in fact an infinite number of bases; to form one, we simply need two orthogonal vectors. Recall, from the appendix on linear algebra, that the eigenvectors of both Hermitian and unitary matrices form a basis for the vector space. Due to this property, we can be sure that the eigenstates of the X-gate and the Y-gate indeed form a basis for 1-qubit states.\n",
```
frankharkins, line 322:
```suggestion
    "We leave it as an exercise to calculate these. There are in fact an infinite number of bases; to form one, we simply need two orthogonal vectors. The eigenvectors of both Hermitian and unitary matrices form a basis for the vector space. Due to this property, we can be sure that the eigenstates of the X-gate and the Y-gate indeed form a basis for 1-qubit states (read more about this in the [linear algebra page](https://qiskit.org/textbook/ch-appendix/linear_algebra.html#Matrices-and-Matrix-Operations) in the appendix)\n",
```

We don't require people read the LA page beforehand so I think better to include it as a reference.

frankharkins (changes requested):
Looks good, thank! Just one small suggestion.

frankharkins (approved):
Thank you!

dhruvbhq:
Thank you @frankharkins ! It was great to work on this issue.

PR #1333, opened 2021-11-30
[4.1.3] switch cost and mixing unitary

marwahaha:
shouldn't the 2-qubit operations come first?


divshacker (commented):
Yup, you are right, the order was incorrect. 
Thanks for the fix.

frankharkins (approved):
LGTM, thank you!

PR #1337, opened 2021-12-06
[3.12] fix encode_message character validation

Techmaster21:
# Changes made
Fixed the encode_message function character validation for superdense coding

# Justification
Previously the function claimed to validate characters in the comments but actually did nothing meaningful.


frankharkins (approved):
Thank you!

PR #1340, opened 2021-12-11
[2.3] swapped control & target qubit in equations

infacc:
# Changes made
Swapped control & target qubit of CNOTs in the text of chapter 2.3.1

# Justification
The control & target qubit of the CNOTs in the text of chapter 2.3.1 were the wrong way around.
The context/code describes the [CNOT gate acting on the state `|+->`][1] but the the state `|-+>` is used in equations.

 [1]: https://en.wikipedia.org/wiki/Controlled_NOT_gate#Details_of_the_computation


infacc:
I just realized, the text is correct, since it assumes little-endianness in the equations. Meaning, the first qubit is the target qubit and the second qubit is the control qubit, which is a bit counterintuitive.

PR #1342, opened 2021-12-14
[8.1] Fixed typo in japanese version.

azennto:
# Changes made

Fixed a typo in the article "Linear Algebra" in the Japanese version.

スクリーンショット 2021-12-14 13 25 49

The "体 F"(field F) in another parts is in bold, but this is the only one that is not.

# Justification

The difference in notation may confuse the reader.


kifumi (approved):
Thank you very much @azennto .  I'm sorry for my late reply.

azennto:
I forgot to build and deploy. I'm sorry. 
Thank you for doing them instead.

PR #1348, opened 2021-12-21
[5.5] Correction in vector presentation of the state

Hosseinberg:
# Changes made
Changes the row vector presentation of psi to column vector presentation. 

# Justification
To be consistent with the rest of the document, a column vector should be used to present the state. Otherwise, It would be confusing.


frankharkins:
@Hosseinberg thanks for taking a look at this. The code cell above:
```
psi_AB = qi.Statevector.from_instruction(qc_AB)
psi_AB.draw('latex', prefix='|\\psi_{AB}\\rangle = ')
```
generates that latex, so if we edited it then the user would not be able to reproduce the output. I agree it could be confusing that the statevector looks like a row vector but should be a column vector.

Qiskit displays `Statevector`s like this because it's essentially a numpy array with extra methods, and it wants to be consistent with that. Maybe solution is to add a line explaining this instead?

PR #1356, opened 2022-01-07
Eigenvectors are not compatible, and must have 1/sqrt(2) multiplications.

soronzonbold:
Eigenvectors should have \frac{1}{\sqrt{2}} multiplication, otherwise \ket{b} is not equal to \ket{0} unless normalized again.

# Changes made

# Justification


frankharkins (approved):
Thank you!

PR #1359, opened 2022-01-10
[1.5] Fix Japanese in case-for-quantum.ipynb

derwind:
# Changes made

- Modify confusing texts

# Justification

- Some Japanese seemed confusing and unclear, so I fixed them.


kifumi (approved):
@derwind thank you!  Sorry for my late reply.

PR #1380, opened 2022-02-16
6.2 Changed to Builder syntax with Parameter class

kUmezawa:
# Changes made

Changed how to write pulse schedule to Builder syntax with Parameter class

# Justification


nkanazawa1989 (changes requested):
Thank @kUmezawa this is great start. Could you please update the execution model to use pulse gate rather than schedule which is currently discouraged. Something like this works:

```python
from qiskit.circuit import QuantumCircuit, Gate

freq = Parameter("freq")

with build(backend) as spect_sched:
    set_frequency(freq, drive_channel(qubit))
    ....

spect = Gate("spect", 1, [freq])

qc = QuantumCircuit(1, 1)
qc.append(spect, [0])
qc.measure(0, 0)
qc.add_calibration(spect, (0,), spect_sched, [freq])

exp_circs = [qc.assign_parameters({freq: f}) for f in freqs]
```
Note: need to check if set_freq command work with pulse gate.

Also now you can remove the first half of this chapter since this is duplicated from what you have done in the previous chapter. If reader is just interested in investigating the higher energy dynamics, one can leverage backend-provided X01 gate to jump start.

### In cell1

Remove these lines
```python
from qiskit.compiler import assemble
from qiskit.pulse.library import Waveform
```

### In cell2

What is it for?
```python
import warnings
warnings.filterwarnings('ignore')
```

This is not necessary if I'm not mistaken. All backends can support pulse experiment through pulse gate for single qubit gate (two qubits are only available in premium systems).
```python
backend_config = backend.configuration()
assert backend_config.open_pulse, "Backend doesn't support Pulse"
```

### In cell35

This can be replaced with `pulse.x(qubit)`
```python
        pulse.call(pi_pulse_01)
```
or, more efficiently, you can use circuit model
```
qc.x(qubit)
qc.append(spect, [0])
```
then scan frequency around 1-2 transition.

kUmezawa:
@nkanazawa1989 
Thank you for your review. I will fix it.

kUmezawa:
@nkanazawa1989 

>Note: need to check if set_freq command work with pulse gate.
It seems set_freq command work with pulse gate
image

kUmezawa:
@nkanazawa1989 

>In cell1
>Remove these lines
OK I'll remove these lines.

>In cell2
>What is it for?
It exists in the original textbook. 
I think this sentence is because the next Warning "ComplexWarning: Casting complex values to real discards the imaginary part" is not displayed in the frequency experiment. 


>This is not necessary if I'm not mistaken. All backends can support pulse experiment through pulse gate for single qubit gate >(two qubits are only available in premium systems).
OK I'll remove these lines.


>Also now you can remove the first half of this chapter since this is duplicated from what you have done in the previous >chapter. If reader is just interested in investigating the higher energy dynamics, one can leverage backend-provided X01 gate >to jump start.
OK I'll remove these chapter.
1. Discriminating the 0 and 1 states
    1.A 0->1 Frequency Sweep
    1.B 0->1 Rabi Experiment
    1.C Build the 0,1 discriminator

>This can be replaced with pulse.x(qubit)
OK I'll change it.

nkanazawa1989:
Thanks @kUmezawa this looks great. I'm happy to see eventually the pulse gate model is working in this notebook!

1. Double check if you remove all redundant imports for reader to easily track the code. for example, 
```python
#import qiskit.pulse as pulse
import qiskit.pulse.library as pulse_lib
```
are not used.

2. It would be better to extract anharmonicity info from the backend in cell2 (in addition to default_qubit_freq) rather than hardcoding `sweep_freqs = np.linspace(-0.38*GHz, -0.32*GHz, num_freqs)`. At the time of writing this, we didn't have anharmonicity value in backend.

3. You don't need to set beta parameter (or DRAG) for X12 spectroscopy. DRAG is used to correct Stark shift from higher levels and the energy level configuration is different in 0-1 and 1-2 subspace. Thus it doesn't make sense to reuse this parameter.

4. If you want to copy the parameter, this would be much cleaner.
```python 
t0, pi_pulse = backend_defaults.instruction_schedule_map.get('x', 0).instructions[0]
pi_params = pi_pulse.pulse.parameters
pi_params.pop("beta")
...

amp = Parameter("amp")
pi_params["amp"] = amp

play(pulse.Gaussian(**pi_params), d0)
```
However I think you don't need to write complicated logic to copy. Since duration and sigma are kind of fixed parameters in our system, you can simply hardcode

```python
# these are typical parameters in single qubit drive in IBM devices
x12_duration = 160
x12_sigma = 40
```

I think this is much readable. Probably you can use the same pulse gate for spectroscopy experiment, since performing spectroscopy with wrong beta parameter may result in calibration error (and also note that we often use long/weaker pulse for spectroscopy to avoid broadening).

5. Since we only have single section now, you can remove `1.` from `1. Discriminating the |0>, |1>  and |2> states`. Same for subsections and references.

kUmezawa:
@nkanazawa1989 
>Double check if you have removed all redundant imports for reader to easily track the code. for example,
Oh, sorry . I removed unnecessary these code
```
from scipy.signal import find_peaks
#import qiskit.pulse as pulse
import qiskit.pulse.library as pulse_lib
```

>It would be better to extract anharmonicity info from the backend in cell2
OK. I added the code.

>You don't need to set beta parameter 
OK . I  use Gaussian pulse in frequency sweep and amplitude sweep.

>you can simply hardcode
OK I adopted a hard code.

>ince we only have single section now, you can remove 1. from 1. Discriminating the |0>, |1> and |2> states. Same for subsections and references.
I changed it like this.
image

kUmezawa:
@nkanazawa1989 
1
The cell4's constant is not used, so I deleted it.
2
I removed the part where warning is disabled.
3
I changed the wording to "there are pulse parameters of the single qubit drive in ibmq_armonk".
4
I fixed the internal logic of get_job_data to use np.real.

nkanazawa1989 (approved):
Thanks @kUmezawa this looks good to me. This chapter will be nicely cleaned up with this PR (replacement of outdated technique, conformation to the pulse gate model,  removal of boilerplate section from previous chapter). Unfortunately the status of hardware doesn't seem to be quite good (seems like DAC power is very small). One thing you could probably do is increasing the spect pulse power, to something around 0.6 (or increasing spect pulse duration) to see if it becomes better, or using another device since now every backend supports experiment with pulse gate of single qubit. If you are willing to follow up in another PR, probably adding Ramsey experiment for fine frequency tuning is another interesting case, especially with the pulse gate model.

frankharkins (commented):
Thank you!

frankharkins (approved):

PR #1381, opened 2022-02-16
[1.4] (ja) Update figure and texts in single-qubit-gates.ipynb

morim3:
# Changes made

update the execution result in single-qubit-gates.ipynb and modified some texts

# Justification

Japanese version of single-qubit-gates.ipynb is partially behind from English version and it was confusing.


kifumi:
@morim3 thank you for correcting the Japanese edition. Could you please update the file on main branch? Master-ja branch is for the new chapters which are not translated yet.

morim3:
@kifumi thanks for the review!

PR #1389, opened 2022-03-01
[1.4] Update japanese translations and some figures in single-qubit-gates.ipynb 

morim3:
# Changes made
update the execution result in single-qubit-gates.ipynb and modified some texts

# Justification
Japanese version of single-qubit-gates.ipynb is partially behind from English version and it was confusing.


kifumi:
@morim3 , hi, thank you for updating the translation. Could you please check below three points?

1) `%config InlineBackend.figure_format = 'svg' # Jupyterノートブックの場合" ` of In [1] does not appear to be in the original code.  Can you please delete it?

2) You are using `qc.draw('mpl')` in the several cells, but  `qc.draw()` is using in the original code.  Can you please change them?

3) Additionally, in "5. The P-gate" section, could you please change from `gate_demo(gates='pauli+h+rz')`  to `gate_demo(gates='pauli+h+p')` ?

morim3:
@kifumi 
Thank you for reviewing. I modified the points.
Could you check if everything looks fine?

kifumi (approved):
Thank you so much.

PR #1396, opened 2022-03-18
6.1 Change the actual machine

kUmezawa:
# Changes made

Change the actual machine from armonk to manila and correct the value of the variable that depends on the actual machine

# Justification

With the end of service of the actual machine armonk, the code will not work as it is now.


kUmezawa:
@nkanazawa1989 @nbronn 

- Change from armonk to manila
- Change the value of drive_sigma_sec, scale_factor and some fitting parameter value
- I get the following error, so add "warnings.filterwarnings ('ignore')"
>447: RuntimeWarning: Dynamic rep rates are supported on this backend. 'rep_delay' will be used instead of 'rep_time'.
  warnings.warn(

nbronn (approved):
Looks good to me, thanks!

frankharkins (approved):
Thank you!

PR #1407, opened 2022-04-21
[3.6] The Japanese translation of the exercise was incorrect.

kUmezawa:
# Changes made

The Japanese translation of the control was missing, so I added it.

# Justification

If the problem statement is different, it will not be a problem, so it needs to be corrected.


kUmezawa:
@kifumi 
Please review this PR. Thank you.

kifumi (approved):
It looks good. Thank you @kUmezawa !

PR #1411, opened 2022-05-03
[1.4] Fix Japanese in single-qubit-gates.ipynb

derwind:
# Changes made

- Modify confusing texts

# Justification

- Some Japanese seemed confusing, so I fixed them: the gates HZH act on the Z-basis rather than the Z-basis acting on the gates.


kifumi (approved):
Thank you! It's good!

PR #1413, opened 2022-05-03
[1.5] Corrected the scope of Einstein's statement.

derwind:
# Changes made

- Modify confusing texts

# Justification

- Corrected Einstein's statement to exclude what Einstein did not say.
- Modified translation of "the other” to clarify texts. "どちらか" is not an appropriate translation.
- Added missing translation of "In further chapters we will explore the consequences of this unique nature."


kifumi (approved):
Thank you so much!

PR #1415, opened 2022-05-04
[2.2] Update and add new translations

derwind:
# Changes made

- Updated translations to correspond to the English version.
- Added a translations that were missing from the existing translations.

# Justification

- Some paragraphs don't correspond to the [English version](https://github.com/qiskit-community/qiskit-textbook/blob/main/content/ch-gates/multiple-qubits-entangled-states.ipynb), especially ones related to `Aer.get_backend('aer_simulator')`.
- Translations for "3.3 Visualizing Entangled States" are missing.


kifumi:
@derwind Thank you so much for updating the translations of many files!  I will check this PR carefully because it includes new section that has never been translated before.  Please wait a moment.

kifumi:
@derwind Thank you so much for updating the translations of many files!  I will check this PR carefully because it includes new section that has never been translated before.  Please wait a moment.

kifumi:
@derwind Sorry for my late answer.  Could you please use "Q-sphere" for translation of "Q-sphere", not "Q球"? There are two places. I discussed about this wording with my colleague and I thought it would be better to use the same "Q-sphere", because we usually call it "Q-sphere".

derwind:
@kifumi Thank you for your suggestion, and I will correct them to use "Q-sphere". I have checked several related Japanese books, and it seems that Bloch sphere is translated as "ブロッホ球", while "Q-sphere" is still "Q-sphere".

kifumi (approved):
Thank you!

PR #1421, opened 2022-05-08
[2.6] Fix codes and one related figure

derwind:
# Changes made

- Fix some codes
- Update one related figure in the output

# Justification

- Reading the description just before the corrected section, it appeared that the order of application of the gates in the code was incorrect


frankharkins (approved):
Thank you!

Issue #1432, opened 2022-05-28
Calibrating qubits using Qiskit Pulse does not work as intended


Yuval-Toren:
When I run the notebook there are two main things that do not work as intended:
1. manila does not support qiskit pulse, only Armonk works with it.
2. instead of getting the expected results I get what seems to be just noise as can be seen here: https://imgur.com/a/o8gUpTT

What can be done to fix this? thanks

HuangJunye:
Having pulse access or not depends on your provider. The open provider only gives pulse access to Armonk. For other devices you can use [pulse gates](https://qiskit.org/documentation/tutorials/circuits_advanced/05_pulse_gates.html) That being said with Armonk getting sunset, we might want to update to notebook to use pulse gates so that all users can still run the same notebook with other systems. @frankharkins What do you think?

kUmezawa:
@HuangJunye @frankharkins Cc @nkanazawa1989 
6-2 has been changed to manila and changed to Pulse Gate.
6-1 has not been changed to Pulse Gate, so I will fix it.
Similarly, modify 6-5 and 6-6 in the same way.

kUmezawa:
@HuangJunye 
Yes, of course, Thank you.

HuangJunye:
I realized I do not have permission to assign. @frankharkins Could you please assign the issue to @kUmezawa ?

PR #1436, opened 2022-06-06
6.1 Modified to use Pulse Gate

kUmezawa:
# Changes made

Modified to use Pulse Gate.

# Justification

Relating issue: https://github.com/qiskit-community/qiskit-textbook/issues/1432

In order to experiment with pulses with manila device using the open provider, we have to write Pulse Gate.


kUmezawa:
@nkanazawa1989 
I changed to use Pulse Gate. Please check it.

nkanazawa1989 (commented):
Thanks @kUmezawa this helps us a lot with migration from schedule to circuit execution model. Here are minor suggestions to use quantum circuit more effectively.

In 3.2:
You don't need to define `exc_gate` since this is X gate. You can write
```python
qc_exc = QuantumCircuit(1, 1)
qc_exc.x(0)
qc_exc.measure(0, 0)
qc_exc.add_calibration("x", (0,), exc_schedule, [])
```

In 3.3:
Same with above. So you can write
```python
qc_t1 = QuantumCircuit(1, 1)

qc_t1.x(0)
qc_t1.delay(delay, 0)
qc_t1.measure(0, 0)
qc_t1.add_calibration("x", (0,), t1_schedule, [])
```

In 3.3:
Here the situation is much more complicated. Because frequency you set in the gate calibration doesn't propagate through the circuit. Since Ramsey experiment is sensitive to frequency offset during the delay, the delay must have frequency shift command, i.e. delay should be defined as a pulse program. Considering this complexity, it would be simpler to define the entire sequence as a single gate, which you are currently doing. I think 3.3 is okey as it is. Once we implement a mechanism to update channel frequency from the circuit, we can decompose, e.g., `ramsey_gate` into two `QuantumCircuit.sx` with `QuantumCircuit.delay` parametrized with delay.

kUmezawa:
@nkanazawa1989 
Added a description of timing constraints.
Reflected the content of the review.

Please check it.

content/ch-quantum-hardware/calibrating-qubits-pulse.ipynb
```diff
@@ -1135,10 +1256,23 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "qc_t1 = QuantumCircuit(1, 1)\n",
     "\n",
-    "t1_schedules = [t1_schedule.assign_parameters({delay: get_dt_from(d)}, inplace=False) for d in delay_times_sec]"
+    "qc_t1.x(0)\n",
+    "qc_t1.delay(delay, 0)\n",
+    "qc_t1.measure(0, 0)\n",
+    "qc_t1.add_calibration(\"x\", (0,), t1_schedule, [delay])\n",
```
nkanazawa1989, line 1273: Here you attached `t1_schedule` to `x` gate. This is a bug, because `t1_schedule` contains delay while the program written in circuit also contains the same delay, so you have 2 delays in total. What you can do here is
```python
qc_t1.add_calibration("x", (0,), pi_pulse)
```

nkanazawa1989 (changes requested):
Thanks @kUmezawa this is awesome! Now user can learn how to manage constraints values reported by backends. Here is a mistake. I'll approve after this is fixed.

kUmezawa:
@nkanazawa1989 
Thank you for your review.
I changed from t1_schedule to pi_pulse.
I removed unnecessary t1_schedule.

nkanazawa1989 (approved):
LGTM! Thanks :)

PR #1444, opened 2022-06-23
[Your First Quantum Game] Add optional feature to game engine

quantumjim:
# Changes made
An option was added to the `QiskitGameEngine` to allow the screen to be interactive. This was then documented in the relevant page (which isn't actually listed within the textbook).

# Justification

Changes required for content in the QEC Summer School.


frankharkins (approved):
Thanks

PR #1461, opened 2022-07-29
[3.8] swapped scalar 1 by identity matrix in grover.ipynb

HannesGitH:
[3.8] use `I` instead of `1` for the identity matrix (since subtracting a scalar from a matrix is undefined)

# Changes made
Swapped `\\mathbb{1}` with `I` in following eqaution: 
```diff
- U = 2 |s> <s| - 1
+ U = 2 |s> <s| - I
```

# Justification
`2 |s> <s|` is a matrix, and its mathematically not clear/defined how to substract a scalar (`1`) from this matrix, using `I` which is commonly used as the identity matrix is better


frankharkins (approved):
Thanks!

PR #1467, opened 2022-08-26
[2.5] Fixed expression result to identity matrix not scalar

giacomoRanieri:
# Changes made

Updated equivalence between the product of a unitary matrix and its Hermitian conjugate

# Justification

The product of a unitary matrix and its Hermitian conjugate is the identity matrix (I), not a scalar (1)


giacomoRanieri:
I've found the same error again, I'm closing this request to open a new one with both the fixes

PR #1488, opened 2022-09-20
update get_expectation

ozlemsalehi:
get_expectation function does not use p, it can be removed

# Changes made

# Justification


frankharkins (approved):
Thank you!

PR #1494, opened 2022-10-07
copy vqe-molecules.ipynb from platypus, because current versions uses aqua

1ucian0:
Based on https://quantumcomputing.stackexchange.com/q/28448/1859, users are still using the VQE chapter that refers to aqua. This PR copies https://github.com/Qiskit/platypus/blob/main/notebooks/ch-applications/vqe-molecules.ipynb into this textbook.


frankharkins (approved):
Thank you

PR #1497, opened 2022-10-10
[4.1.3] reverse bitstrings

### LINKED ISSUE #1495: 4.1.3 (QAOA-MaxCut) Bit strings need to reversed while computing cut sizes
### Describe the issue

This pertains to the `compute_expectation` and `maxcut_obj` functions in the tutorial. 

Qiskit uses the little endian format. Hence, the bit-strings in the `counts` dictionary that gets inputted to the `compute_expectation` is not in the same order as is expected by the `maxcut_obj` function. Hence, the bitstrings need to be reversed. One way to do this is by adding the string reversal operation in the `compute_expectation` function, as done below:

`
for bitstring, count in counts.items():
        bitstring = bitstring[::-1]
        obj = maxcut_obj(bitstring, G)
        avg += obj * count
        sum_count += count
        
    return avg/sum_count
`

### Describe what you expect to see

Correct implementation.

Note that even with the current, buggy version, QAOA finds angles that result in a high probability of finding the solution (as can be seen from the histogram). However, testing this code on larger graphs reveals extremely poor performance (with the buggy version), but a substantially improved result with the bug-free version proposed above.

PratikSathe:
# Changes made
Reverse bit-strings before evaluating cut sizes

# Justification
Bit-strings in the counts dictionary are in little-endian format. However, input expected by the cut size evaluation function is not.

fixes #1495 

@bernalde

### DIFF
--- a/content/ch-applications/qaoa.ipynb
+++ b/content/ch-applications/qaoa.ipynb
@@ -3443,7 +3443,7 @@
     "    sum_count = 0\n",
     "    for bitstring, count in counts.items():\n",
     "        \n",
-    "        obj = maxcut_obj(bitstring, G)\n",
+    "        obj = maxcut_obj(bitstring[::-1], G)\n",
     "        avg += obj * count\n",
     "        sum_count += count\n",
     "        \n",


frankharkins (approved):
Thank you!

PR #1514, opened 2022-11-14
Added missing images for Japanese edition

kifumi:
# Changes made

I added two missing images for Japanese edition of[ Quantum walk algorithm](https://qiskit.org/textbook/ja/ch-algorithms/quantum-walk-search-algorithm.html).
# Justification


frankharkins (approved):
Thank you!

Issue #1515, opened 2022-11-16
[Chapter-3 (Quantum Protocols and Quantum Algorithms)] Sec 3.8  Grover's Algorithm


MonitSharma:
### Describe the issue

In the Example: 2 qubit:  Reflections paragraph, it says 
>>Reflection  U_s. In order to complete the circuit we need to implement the additional reflection  Us=2|s⟩⟨s|−I . Since this is a reflection about |s⟩ , we want to add a negative phase to every state orthogonal to |s⟩ .

It's not a reflection about |s⟩ , it's a reflection about |s'⟩ .

### Describe what you expect to see

The paragraph should be changed to :
>>Reflection  U_s. In order to complete the circuit we need to implement the additional reflection  Us=2|s⟩⟨s|−I . Since this is a reflection about |s'⟩ , we want to add a negative phase to every state orthogonal to |s⟩ .

frankharkins:
I think the original is correct. There are two reflections: the oracle reflects around |s'>, and the diffuser reflects around |s>. Since this step is referring to the diffuser I think it should stay as |s>?

![grover_step3](https://user-images.githubusercontent.com/36071638/202440102-d3400a99-6fb3-4687-8544-1ebc62247749.jpg)

MonitSharma:
Yes. You are correct. I misinterpreted it as the first rotation.   Closing this issue.
Thanks for clearing the doubt.

frankharkins:
No problem, thank you!

PR #1516, opened 2022-11-16
Chapter [3] Section [3.8] Fixed the error of |s> to |s'>

MonitSharma:
# Changes made
Addressed the Issue #1515 

Changed the "reflections about |s>" to "reflections about |s'>" and also the "perpendicular to |s>" to "perpendicular to |s'>"

# Justification

In the grover's algorithm, the two orthogonal states are the winner state |w> and the superposition state excluding the winner state i.e |s'>. 
So the reflection is of the superposition state |s> about the perpendicular state |s'>.


frankharkins:
@MonitSharma it looks like the notebook is not valid, can you reset your changes and try again please?

MonitSharma:
Not able to reopen the file. Anytime I propose the change the file just crashes and returns an Error of the page not being found.
[This](https://github.com/qiskit-community/qiskit-textbook/blob/main/content/ch-algorithms/grover.ipynb) is the file I'm trying to correct.  Made another PR

Issue #1512, opened 2022-11-12
[What is Quantum?] off-by-one in coin toss simulator


RahulGopan0:
### Describe the issue

While tossing the coin with the simulator, the initial no: of coin/s is set to 0, which is obscured, and the simulator goes on to toss '0' coin/s and still gives a result, which in reality should not be the case. This is the [link](https://qiskit.org/textbook/what-is-quantum.html).

### Describe what you expect to see

It would be better if the initial no: of coin/s is set to 1 rather than 0. This should also automatically solve the issue of running the simulator with 0 coin/s.

Hoping to see the issue fix.

1ucian0:
There seems to be an off-by-one in the coin toss simulator in https://qiskit.org/textbook/what-is-quantum.html#Experiment-#1:-Single-Coin-Toss

![Screenshot 2022-11-13 at 18 17 15](https://user-images.githubusercontent.com/766693/201534957-f5e5cc39-3dd5-4120-9763-9de785927bde.png)

When the screen loads, it's in 0, which actually means 1. if you put the dial into 1, it simulates twice.

Tagging as a `good first issue`.

RahulGopan0:
>

PR #1517, opened 2022-11-16
[What is Quantum?] off-by-one in coin toss simulator

MonitSharma:
# Changes made

Changed the value of the `value` attribute of `repeat_slider` function in the "Experiment 1" from `0` to `1`.

# Justification

Fixed the issue #1512 
>>While tossing the coin with the simulator, the initial no: of coin/s is set to 0, which is obscured, and the simulator goes on to toss '0' coin/s and still gives a result, which in reality should not be the case.  

Changed it to be 1.


frankharkins (approved):
Thank you!

PR #1528, opened 2022-11-24
[6.3] Fixed operation of annihilation operator

dryicecube:
--->
# Changes made
$\sigma ^ + \ket{1}=\ket{0}$ was incorrect.
Changed the operation of and creation operator in section 6.  

# Justification

The $\sigma^+$ is a creation operator, but for a two-level system, $\sigma ^ + \ket{1}=\text{NULL}$. The actual operation is $\sigma ^ - \ket{1}=\ket{0}$ i.e. annihilation operator takes the system from $\ket{1}$ to $\ket{0}$


nbronn:
Thanks for fixing the error... LGTM!

frankharkins (approved):
Thank you!

Issue #1529, opened 2022-11-24
Representing Qubit States: Section 3.2 - Behaviour of plot_bloch_vector() in the legacy textbook no longer consistent with qiskit


mogul27:
## Describe the Issue

In Section 3.2 of Representing Qubit States, the legacy textbook states that plot_bloch_vector() does not accept spherical coordinates. However, in more recent versions of qiskit (whilst cartesian coordinates are the default behaviour) spherical coordinates can be passed to plot_bloch_vector() with an argument specifying that spherical coordinates are being passed.

## Expected Solution

Explanation of plot_bloch_sphere() behaviour in this section of the textbook should be updated to reflect this behaviour and the plot_bloch_vector_spherical() function in the textbook may now be redundant.

## Anything else we should know?

Link to docs showing behaviour of plot_bloch_vector
https://qiskit.org/documentation/stubs/qiskit.visualization.plot_bloch_vector.html?highlight=plot_bloch_vector#qiskit.visualization.plot_bloch_vector

PR #1530, opened 2022-11-24
[3.2] Changed plot_bloch_sphere() function behaviour explanation to reflect the current behaviour.

mogul27:
# Changes made

In Section 3.2 of Representing Qubit States, the legacy textbook currently states that plot_bloch_vector() does not accept spherical coordinates. This explanation has been changed to reflect that in more recent versions of qiskit, (whilst cartesian coordinates are the default behaviour) spherical coordinates can be passed to plot_bloch_vector() with an argument specifying that spherical coordinates are being passed. 

The new text reads as follows: (Qiskit has a function to plot a bloch sphere, `plot_bloch_vector()`. This function accepts cartesian coordinates by default, so we have included a function that does the conversion automatically. However, `plot_bloch_vector()` can also accept spherical coordinates by passing the argument `coord_type='spherical'`.)

The issue is also described in issue #1529.

# Justification

The current explanation of plot_bloch_sphere() is out of date and does not describe the actual behaviour of the function. This could mislead those new to Qiskit into thinking that they are unable to use spherical coordinates with plot_bloch_sphere(). 

The documentation for the function plot_bloch_sphere() illustrating the correct behaviour can be found at https://qiskit.org/documentation/stubs/qiskit.visualization.plot_bloch_vector.html?highlight=plot_bloch_vector#qiskit.visualization.plot_bloch_vector


mogul27:
Changes requested from frankharkins have now been addressed commit a542505.

frankharkins (approved):
Thank you!

PR #1543, opened 2022-12-24
[1.4] Article change for grammatical correctness: changed 'an' to 'a'. 

davisclarke:
# Changes made
Changed article in "The T-gate is a very commonly used gate, it is an P-gate with $\\phi = \\pi/4$" to 
"The T-gate is a very commonly used gate, it is a P-gate with $\\phi = \\pi/4$"

# Justification
"an P-gate" should be "a P-gate" to agree with English article rules. Since the starting letter, 'p-', in p-gate (phase gate) is a consonant that is not silent, 'a' would be the correct article.


frankharkins (approved):
Thanks!

PR #1574, opened 2023-04-12
add client redirects

vabarbosa:
# Changes made

add client side redirection for each of the pages in the Textbook to the corresponding learning.qiskit.org page

# Justification

this PR is part of a bigger set of changes needed to redirect the Textbook to learn.qiskit.org (and only a temporary approach while server side redirects is being set up)


vabarbosa:
closing.... no longer needed

PR #1576, opened 2023-04-16
[Single Qubit Gates] 4 - Digression: Measurement in Different Bases: Added Clarification (Resolves #1575)

### LINKED ISSUE #1575: [Single Qubit Gates] 4 - Digression: Measurement in Different Bases: Additional Clarification needed
### Is your feature request related to a problem? Please describe.

In the chapter **Single Qubit Gates**, in section 4 : **Digression: Measurement in Different Bases**,  there is definitely some additional clarification needed. 

problem

These are the problems I see -

1. When the book states, 'In the quick exercises above, we saw you could create an X-gate by sandwiching out Z-gate between two H-gates' - The statement is false because no exercises before this line discuss this.

2. When the book states, 'Following the same logic, we have created....'  - The statement seems to be a bit lousy and ends up confusing the concepts and the point the Section 4 is trying to make.

3. Then the book hastily goes on state - 'Since the process of measuring can have different effects depending on the ...' - This statement has no direct relationship to the concepts being discussed above. And yet the statement seems to be in the same paragraph as the previous statement. This confuses the reader further.

4. In addition- All the 3 problems prevent the section from having a logical flow and the section looks a bit confusing due to that.

### Describe the solution you'd like

I have reordered a couple lines and added an extra line for additional clarification. To me, the section now looks complete, extremely clear, and achieves the intended logical flow.

I have added these changes as a PR. 
[DevenPanchal:add-Deven-Panchal](https://github.com/DevenPanchal/qiskit-textbook/tree/add-Deven-Panchal)

I am adding an image to see the changes quickly.
solution

### Describe alternatives you've considered

None. Not required.

### Additional context

None

DevenPanchal:
Add clarification, improve subject matter flow
Resolves #1575 

# Changes made
I have reordered a couple lines and added an extra line for additional clarification. 

# Justification
I was seeing problems in the logical flow and clarity causing the section to be overly confusing.
I have detailed the problems in Issue #1575


DevenPanchal:
Recognized a minor mistake. So closing this one. Will submit a new one.

